PR #1, opened 2019-03-11
Start of repo and AttachResult component

mikegron:
- Updated readme and other files, starting from search-ui-seed.
- Added AttachResult component and tests.

We want this repository to be home to generic and reusable components which could be used by any integrations in the future. So any comment is appreciated!

I started from [search-ui-seed](https://github.com/coveo/search-ui-seed) and added a basic, generic AttachResult component. The component is super simple and leverages Promises, it's basically a toggle button which also sends analytics events. You just have to give it 3 functions: attach, detach and optionally isAttached. These are configured just like any other component options, there is an example in index.html.

image image image


src/ui/AttachResult/AttachResult.ts
```diff
@@ -0,0 +1,180 @@
...
+        detach: ComponentOptions.buildCustomOption(name => () => new Promise<void>((resolve, reject) => { resolve(); }), {
+            defaultFunction: () => () => new Promise<void>((resolve, reject) => { resolve(); })
+        })
+    };
+
+    constructor(
+        public element: HTMLElement,
+        public options: IAttachResultOptions,
+        public bindings?: IComponentBindings,
+        public queryResult?: IQueryResult
```
samisayegh: Does queryResult need to be a separate constructor arg, or does it make sense to put it under `options`?

src/ui/AttachResult/AttachResult.ts
```diff
@@ -0,0 +1,180 @@
...
+        this.initialize();
+
+        this.bind.on(this.element, "click", this.toggleAttached);
+    }
+
+    /**
+     * Gets whether or not the result is currently attached.
+     */
+    public isAttached() : boolean {
+        return this.attached;
```
samisayegh: Could this return `undefined`? Would it need to be casted, for example by returning `return !!this.attached`?

src/ui/AttachResult/AttachResult.ts
```diff
@@ -0,0 +1,180 @@
...
+    /**
+     * Attach the query result.
+     */
+    public attach() : Promise<void> {
+        if (this.attached || this.loading) {
+            return Promise.resolve();
+        }
+
+        this.setLoading(true);
+        return this.options.attach(this.queryResult)
```
samisayegh: I would recommend using `async/await` syntax rather than chaining `.then`

src/ui/AttachResult/AttachResult.ts
```diff
@@ -0,0 +1,180 @@
...
+            .then(() => { this.attached = true; })
+            .then(() => { 
+                let customData: IAnalyticsCaseAttachMeta = {
```
samisayegh: It would be nice to extract this block as a separate method `logAttachCaseAnalyticsEvent`.

Similar comments for `detach`

src/ui/AttachResult/AttachResult.ts
```diff
@@ -0,0 +1,180 @@
...
+                };
+          
+                this.usageAnalytics.logCustomEvent<IAnalyticsCaseDetachMeta>(analyticsActionCauseList.caseDetach, customData, this.root);
+            })
+            .finally(() => { this.setLoading(false); });
+    }
+
+    /** Toggle the state of the component. If the current result is not attached, attach it, if not, detach it. */
+    public toggleAttached() : void {
+        if (this.attached) {
```
samisayegh: For simple `if/else` logic like this one, I like using ternaries: e.g. 
```
this.attached ? this.detach() : this.attach()
```

src/ui/AttachResult/AttachResult.ts
```diff
@@ -0,0 +1,180 @@
...
+            this.detach();
+        } else {
+            this.attach();
+        }
+    }
+
+    protected initialize() : void {
+        // Adds HTML elements needed to display this component.
```
samisayegh: I would remove this comment. It is better to reserve them for "why" you are doing something rather than "what" you are doing.

src/ui/AttachResult/AttachResult.ts
```diff
@@ -0,0 +1,180 @@
...
+        this.buttonElement = $$("div", {}, PaperclipIcon).el;
+        this.element.appendChild(this.buttonElement);
+
+        this.tooltipElement = $$("div", { className: "coveo-caption-for-icon" }).el;
+        this.element.appendChild(this.tooltipElement);
+
+        this.attached = false;
```
samisayegh: Could this be set on line 35 instead? If so, then my comment on casting using `!!` should no longer be necessary.

src/ui/AttachResult/AttachResult.ts
```diff
@@ -0,0 +1,180 @@
...
+        this.render();
+
+        // Resolve the current result for the component and the initial state.
+        if (this.options.isAttached) {
```
samisayegh: I would extract this logic into a separate method, named something along the lines of `synchronizeAttachedState` or `updateAttachedState`

tests/ui/AttachResult.spec.ts
```diff
@@ -0,0 +1,237 @@
...
+
+        it("should call isAttached on initialize", () => {
+            attachResult = Mock.optionsResultComponentSetup(AttachResult, { isAttached: faker.isAttachedMock }, fakeResult);
+            expect(faker.isAttachedMock).toHaveBeenCalledWith(fakeResult);
+        });
+
+        it("should be attached when isAttached returns true", (done) => {
+            isAttachedSpy.and.returnValue(Promise.resolve(true));
+            attachResult = Mock.optionsResultComponentSetup(AttachResult, { isAttached: faker.isAttachedMock }, fakeResult);
+            setTimeout(() => {
```
samisayegh: Seems strange that these `setTimeout` are needed?

tests/ui/AttachResult.spec.ts
```diff
@@ -0,0 +1,237 @@
...
+                detachCaption: "detach me",
+                attachCaption: "attach me"
+            }, fakeResult);
+
+            attachResult.cmp.attach().then(() => { done(); });
+        });
+
+        it("should only detach if attached", (done) => {
+            attachResult.cmp.detach()
+                .then(() => {
```
samisayegh: You could use async/await here,

src/ui/AttachResult/AttachResult.ts
```diff
@@ -0,0 +1,180 @@
...
+    attachCaption?: string;
+    /** Specifies the tooltip displayed when the result is already attached. */
+    detachCaption?: string;
+    /** The function that is called when the user wants to attach a result. */
+    attach: (queryResult: IQueryResult) => Promise<void>;
+    /** The function called when the user wants to un-link a result. */
+    detach: (queryResult: IQueryResult) => Promise<void>;
+    /** Optional function to check the initial state of the component. */
+    isAttached?: (queryResult: IQueryResult) => Promise<boolean>;
+}
```
ThibodeauJF: If you ever want to provide an option for tooltip positioning, instead of it always being at the bottom, for the Quickview component we use PopperJs
https://github.com/coveo/search-ui/blob/master/src/ui/Quickview/Quickview.ts

tsconfig.json
```diff
@@ -0,0 +1,22 @@
+{
+  "compilerOptions": {
+    "module": "commonjs",
+    "target": "es5",
+    "noImplicitAny": false,
```
samisayegh: Since this is a young project, this would be a good option to set to `true`.

Other options that are nice: `noUnusedLocals`, `noImplicitReturns`

tslint.json
```diff
@@ -0,0 +1,113 @@
+{
```
samisayegh: I recommend removing this config and using the opinionated linter [prettier ](https://www.npmjs.com/package/prettier) instead. As you prefer,

src/ui/AttachResult/AttachResult.ts
erocheleau: What do you usually do with your components that require a result?

src/ui/AttachResult/AttachResult.ts
```diff
@@ -0,0 +1,180 @@
...
+        this.loading = loading;
+        this.render();
+    }
+
+    protected render() : void {
+        $$(this.buttonElement).removeClass("coveo-icon-attached");
+        $$(this.buttonElement).removeClass("coveo-icon-attach");
+        $$(this.buttonElement).removeClass("coveo-icon-loading");
+
+        if (this.loading) {
```
erocheleau: You could extract some of that logic into a private method: getClassFromState or something of the sort that would return either "coveo-icon-loading" or "coveo-icon-attached" or "coveo-icon-attach".

tests/ui/AttachResult.spec.ts
erocheleau: To simulate the back and forth of a server call?

LICENSE
```diff
@@ -1,201 +1,13 @@
-                                 Apache License
-                           Version 2.0, January 2004
-                        http://www.apache.org/licenses/
+Copyright 2017 Coveo
```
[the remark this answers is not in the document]
mikegron: yes it was asked by the sec team!

pages/index.html
```diff
@@ -0,0 +1,117 @@
...
+    <link rel="stylesheet" href="./css/CoveoFullSearch.css"/>
+    <link rel="stylesheet" href="./css/CoveoJsSearchExtensions.css"/>
+    <script src="http://localhost:8080/js/CoveoJsSearch.Lazy.js"></script>
+    <script src="js/CoveoJsSearchExtensions.js"></script>
+    <script>
+      document.addEventListener('DOMContentLoaded', function () {
+        Coveo.SearchEndpoint.configureSampleEndpointV2();
+        Coveo.init(document.body, {
+          AttachResult: {
+            attach: function(id) { return new Promise((resolve, reject) => { console.log("attached " + id); setTimeout(resolve, 1000);} ); },
```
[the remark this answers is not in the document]
mikegron: good point!

sass/AttachResult.scss
```diff
@@ -0,0 +1,71 @@
+$attachResult_grey: #adadad;
```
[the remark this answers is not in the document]
mikegron: I like this idea 👍

src/ui/AttachResult/AttachResult.ts
mikegron: What i've seen is normally it's in the constructor but i'm not 100% why

tests/ui/AttachResult.spec.ts
mikegron: i was having a hard time trying to make the test work, for some reason I cannot call `done()` in the spy `.and.callFake`, the test kept timeouting.. do you have an idea why?

src/ui/AttachResult/AttachResult.ts
samisayegh: I see the `ChatterLikedBy` component in the search-ui accepts the `IQueryResult` as a fourth constructor param. What you did is consistent 🤷‍♂️

tests/ui/AttachResult.spec.ts
samisayegh: Hmm, I've run into problems when using promises with the jest framework. There I solve it by moving the async method into the `beforeEach`. If the promise is the issue, maybe there is a thread that discusses it for Jasmine?

tests/ui/AttachResult.spec.ts
mikegron: that's a lot cleaner :+1:

tsconfig.json
mikegron: nice I didn't know these options

src/ui/AttachResult/AttachResult.ts
mikegron: it's already kind of small and i still have to add the text so not sure it's worth it

PR #2, opened 2019-04-23
Switch the project to es2015 module

jeremierobert-coveo:
# What was changed ?

I changed configuration files to enable webpack **tree-shaking**.
The goal of this is to ensure that when the project is imported somewhere, only required components will affect the **global scope** instead of the whole library.
I also change the file structure a bit to ease this change.

# Was was tested ?

I checked each task within the `package.json` file


package.json
```diff
@@ -2,16 +2,20 @@
...
+
+    "main": "./bin/js/index.js",
+    "types": "./bin/typings",
+
+    "sideEffects": false,
+    
     "scripts": {
-        "test": "karma start --single-run",
+        "test": "karma start ./config/karma.config.js --single-run",
         "testCoverage": "npm run test && cat ./bin/coverage/lcov.info | node ./node_modules/coveralls/bin/coveralls.js",
```
mikegron: You need to change the path here for `../node_modules/`,  I think that's why coveralls didn't work?

mikegron:
Nice, I like the new file structure a lot better!

package.json
jeremierobert-coveo: Will try but it does not makes sense to me. I mean, the package.json file is on root 😕 .

mikegron: oh you're right.. nevermind then :P

mikegron: ah, make sure the coverage is output in the right folder, in the karma.config.json, it outputs to `./bin/coverage`, pretty sure that will be in the config folder then

jeremierobert-coveo: Oups 😱

PR #3, opened 2019-06-21
Filter Attached Results component

mikegron:
Very simple small component that takes in a function to retrieve an array of uri hashes to add at query time i the avanced expression. In the exemple we are keeping this list up to date in memory but in an integration it would be saved in a database of some sorts. The component uses a Checkbox underneat for the toggling logic and sends analytics automatically.

image


pages/index.html
```diff
@@ -49,7 +57,7 @@
         </div>
         <div class="coveo-main-section">
             <div class="coveo-facet-column">
-                <div class="CoveoAttachedResultsFilter"></div>
```
erocheleau: Are we planning on having more than 1 `CoveoFilter*` ?

src/components/FilterAttachedResults/FilterAttachedResults.ts
```diff
@@ -0,0 +1,149 @@
...
+
+export interface IAnalyticsAttachedResultsFilterMeta {
+    filteredAttachedResults: boolean;
+}
+
+export interface IFilterAttachedResultsOptions {
+    /** Specifies the text displayed next to the checkbox. */
+    text?: string;
+    /** The function that is called when retrieving uri hashes to filter. */
+    getUriHashes?: () => string[];
```
erocheleau: Are we absolutely sure that we will always use the urihashes?
Where has permanentId gone?
Should we let people define their own fields?

src/components/FilterAttachedResults/FilterAttachedResults.ts
```diff
@@ -0,0 +1,149 @@
...
+    Initialization,
+    Checkbox,
+    QueryEvents,
+    IBuildingQueryEventArgs,
+    QueryStateModel,
+    load,
+    IAttributesChangedEventArg
+} from 'coveo-search-ui';
+
+export interface IAnalyticsAttachedResultsFilterMeta {
```
erocheleau: Metadata?

src/components/FilterAttachedResults/FilterAttachedResults.ts
```diff
@@ -0,0 +1,149 @@
...
+ * search only for attached results.
+ */
+export class FilterAttachedResults extends Component {
+    static ID = 'FilterAttachedResults';
+
+    private checkbox: Checkbox;
+
+    static options: IFilterAttachedResultsOptions = {
+        text: ComponentOptions.buildStringOption({
+            defaultValue: 'Filter Attached Results'
```
erocheleau: I could be wrong, but I think in english `To filter something` means to **remove** these things, not "show only attached results".

src/components/FilterAttachedResults/FilterAttachedResults.ts
```diff
@@ -0,0 +1,149 @@
...
+        this.bind.onQueryState(
+            'change:',
+            QueryStateModel.getFacetId(FilterAttachedResults.ID),
+            (args: IAttributesChangedEventArg) =>
+                this.handleQueryStateChange(args)
+        );
+    }
+
+    public isSelected(): boolean {
+        return this.checkbox && this.checkbox.isSelected();
```
erocheleau: Should this "throw" if there's no checkbox?
Or just return false?

src/components/FilterAttachedResults/FilterAttachedResults.ts
```diff
@@ -0,0 +1,149 @@
...
+    public toggle(): void {
+        if (this.isSelected()) {
+            this.checkbox.reset();
+        } else {
+            this.checkbox.select(true);
+        }
+    }
+
+    protected initialize(): void {
+        let mainSection = $$('div', { className: 'CoveoFacet' });
```
erocheleau: const here too

src/components/FilterAttachedResults/FilterAttachedResults.ts
```diff
@@ -0,0 +1,149 @@
...
+        let headerSection = $$('div', { className: 'coveo-facet-header' });
```
erocheleau: const

src/components/FilterAttachedResults/FilterAttachedResults.ts
```diff
@@ -0,0 +1,149 @@
...
+        this.queryStateModel.set(
+            QueryStateModel.getFacetId(FilterAttachedResults.ID),
+            this.checkbox.isSelected()
+        );
+        this.triggerQuery();
+    }
+
+    private triggerQuery() {
+        this.usageAnalytics.logSearchEvent<IAnalyticsAttachedResultsFilterMeta>(
+            { name: 'filterAttachedResults', type: 'misc' },
```
erocheleau: Should be static constants probably

src/components/FilterAttachedResults/FilterAttachedResults.ts
```diff
@@ -0,0 +1,149 @@
...
+        } else {
+            this.checkbox.reset();
+        }
+    }
+
+    private handleBuildingQuery(args: IBuildingQueryEventArgs) {
+        if (this.checkbox.isSelected()) {
+            const uriHashes = this.options.getUriHashes();
+            args.queryBuilder.advancedExpression.add(
+                '@urihash=(' + uriHashes.join(',') + ')'
```
erocheleau: Could probably check if `uriHashes` is empty before adding it to the advancedExpression?

```JavaScript
var uriHashes=[]; 
'@urihash=(' + uriHashes.join(',') + ')'
// "@urihash=()"
```

src/components/FilterAttachedResults/FilterAttachedResults.ts
mikegron: good point, we should let the user decide.. we could add an option to set the field and simply use that when building the query!

pages/index.html
mikegron: maybe, what is your point 📦

src/components/FilterAttachedResults/FilterAttachedResults.ts
mikegron: its the standard to suffix only with Meta !

src/components/FilterAttachedResults/FilterAttachedResults.ts
mikegron: :enfrancaissvp: 👍

src/components/FilterAttachedResults/FilterAttachedResults.ts
mikegron: not really, if there is nothing to show then we want to show nothing

src/components/FilterAttachedResults/FilterAttachedResults.ts
```diff
@@ -0,0 +1,149 @@
...
+
+        this.initialize();
+
+        this.queryStateModel.registerNewAttribute(
+            QueryStateModel.getFacetId(FilterAttachedResults.ID),
+            false
+        );
+        this.bind.onRootElement(
+            QueryEvents.buildingQuery,
+            (args: IBuildingQueryEventArgs) => this.handleBuildingQuery(args)
```
jeremierobert-coveo: this.handleBuildingQuery.bind(this) instead of an arrow ?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Function/bind

src/components/FilterAttachedResults/FilterAttachedResults.ts
```diff
@@ -0,0 +1,149 @@
...
+        );
+        this.bind.onQueryState(
+            'change:',
+            QueryStateModel.getFacetId(FilterAttachedResults.ID),
+            (args: IAttributesChangedEventArg) =>
+                this.handleQueryStateChange(args)
```
jeremierobert-coveo: this.handleQueryStateChange.bind(this) instead of an arrow ?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Function/bind

tests/FilterAttachedResults/FilterAttachedResults.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+import { Mock, Simulate } from 'coveo-search-ui-tests';
+
+describe('FilterAttachedResults', () => {
+    let filter: Mock.IBasicComponentSetup<FilterAttachedResults>;
+
+    beforeEach((done) => {
+        filter = Mock.basicResultComponentSetup<FilterAttachedResults>(
+            FilterAttachedResults
+        );
+        setTimeout(done, 0);
```
jeremierobert-coveo: Can you use clock instead of direcxt setTimeout ?
See https://jasmine.github.io/api/2.6/Clock.html.

tests/FilterAttachedResults/FilterAttachedResults.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+        expect(filter.cmp.element.querySelector('input[type=checkbox]')).not.toBeNull();
+    });
+
+    describe("when setting options", () => {
+        beforeEach((done) => {
+            filter = Mock.optionsComponentSetup<FilterAttachedResults, IFilterAttachedResultsOptions>(
+                FilterAttachedResults,
+                { text: "much rain", getUriHashes: () => ["foo", "bar"] }
+            );
+            setTimeout(done, 0);
```
jeremierobert-coveo: Can you use clock instead of direcxt setTimeout ?
See https://jasmine.github.io/api/2.6/Clock.html.

pages/index.html
mikegron: @erocheleau would it be better to rename it "FilterResults" only or "FieldFilterCheckbox" somethinglike that?

src/Index.ts
```diff
@@ -1,3 +1,6 @@
 // This entry point defines all the components that are included in the extensions.
 
 export { AttachResult } from './components/AttachResult/AttachResult';
+export {
+    FilterAttachedResults
+} from './components/FilterAttachedResults/FilterAttachedResults';
```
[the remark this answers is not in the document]
mikegron: It's Prettier that decides if the line is too long .. i'll change that since 80 is way too small lol

tests/FilterAttachedResults/FilterAttachedResults.spec.ts
mikegron: I used the trick of setTimeout with 0 ms to send it to the end of the stack, do you think using clock().tick(1) would be cleaner?

jeremierobert-coveo: Yeah, this way your test is dettach from the main clock.

src/components/FilterAttachedResults/FilterAttachedResults.ts
erocheleau: I would say, can we have an option to decide which field to be used to uniquely identify results. But default to urihash?

src/components/FilterAttachedResults/FilterAttachedResults.ts
erocheleau: I'm saying, if uriHashes is empty, the query will contain `@urihash=()`.

You should check first if `uriHashes` is empty, and simply don't add anything to the query if it's empty?

mikegron: No that's what i'm saying, if there is notrhing to show we still want to add it to the query so nothing will show in the results, which is correct !

src/components/ResultsFilter/ResultsFilter.ts
```diff
@@ -0,0 +1,134 @@
...
+        });
+
+        this.element.appendChild(mainSection.el);
+    }
+
+    private async createCheckbox(): Promise<Checkbox> {
+        if (Coveo.Checkbox === undefined) {
+            await load('Checkbox');
+        }
+        return new Checkbox(checkbox => this.handleCheckboxChange(), this.options.text);
```
jeremierobert-coveo: bind instead of arrow

config/karma.config.js
```diff
@@ -27,7 +27,7 @@ var configuration = {
     reporters: ['coverage-istanbul', 'spec'],
     coverageIstanbulReporter: {
         dir: path.resolve('./bin/coverage'),
-        reports: ['cobertura', 'json', 'lcov', 'text-summary'],
+        reports: ['html', 'cobertura', 'json', 'lcov', 'text-summary'],
```
louis-bompart: Hmm odd, lcov generate the html one normally: https://istanbul.js.org/docs/advanced/alternative-reporters/

src/components/ResultsFilter/ResultsFilter.ts
```diff
@@ -0,0 +1,134 @@
...
+    Initialization,
+    Checkbox,
+    QueryEvents,
+    IBuildingQueryEventArgs,
+    QueryStateModel,
+    load,
+    IAttributesChangedEventArg
+} from 'coveo-search-ui';
+
+export interface IAnalyticsFilteredResultsMeta {
```
louis-bompart: Document the exported class

src/components/ResultsFilter/ResultsFilter.ts
```diff
@@ -0,0 +1,134 @@
...
+        this.queryStateModel.registerNewAttribute(QueryStateModel.getFacetId(ResultsFilter.ID), false);
+        this.bind.onRootElement(QueryEvents.buildingQuery, this.handleBuildingQuery.bind(this));
+        this.bind.onQueryState(
+            'change:',
+            QueryStateModel.getFacetId(ResultsFilter.ID),
+            this.handleQueryStateChange.bind(this)
+        );
+    }
+
+    public isSelected(): boolean {
```
louis-bompart: Document the public methods

PR #13, opened 2019-07-25
Beta deployment

jeremierobert-coveo:
Add a conditionnal deploy when the tag has beta in it.


jeremierobert-coveo:
It's more simple as our use case is.

louis-bompart:
While working on tagging etc, I'm not convinced by the fact that we seems to manually tag versions to trigger a new release, I think that we should have a beta tag, yes, but this tag should merely point at the tip of the master branch (i.e. latest PR merged) . Same for the latest tag in fact, it could point the tip of a'production' branch. We would update the production branch through a PR and could define the version bump through some tools like [semantic-release](https://github.com/semantic-release/semantic-release) (i.e. if we only have bug fix, bump the patch, if we have a feature, bump the minor, if we have a breaking change, bump the major. semver in short)

PR #14, opened 2019-07-25
Sort out svg shenanigans

louis-bompart:
Simple PR, just noticed that the icon import was flagged in AttachedResult.ts
I chose to use the utils instead of including the ref typings for the svg module.


config/tsconfig.json
```diff
@@ -16,6 +16,6 @@
         "typeRoots": ["../src/typings"],
         "types": ["jasmine", "coveo-search-ui", "coveo-search-ui-tests"]
     },
-    "include": ["../src/**/*.ts", "../src/**/*.svg", "../src/**/*.json"],
+    "include": ["../src/**/*.ts", "../img/**/*.svg", "../src/**/*.json"],
```
jeremierobert-coveo: Je suis currieux de vois le code ressemble a quoi une fois transpiller ?

src/utils/icons.ts
```diff
@@ -5,8 +5,11 @@ import * as DUPLICATE from '../../img/duplicate.svg';
 import * as SEARCH from '../../img/search.svg';
 import * as VIEW from '../../img/view.svg';
 import * as DOT from '../../img/dot.svg';
+import * as PAPER_CLIP from '../../../img/paperclip.svg';
```
jeremierobert-coveo: ../.. et non ../../..

config/tsconfig.json
louis-bompart: `this.buttonElement = Object(coveo_search_ui__WEBPACK_IMPORTED_MODULE_0__["$$"])('div', {}, _utils_icons__WEBPACK_IMPORTED_MODULE_1__["paperclipIcon"]).el;`
`this.buttonElement = Object(coveo_search_ui__WEBPACK_IMPORTED_MODULE_0__["$$"])('div', {}, _img_paperclip_svg__WEBPACK_IMPORTED_MODULE_1__).el;`

The webpack shenanigans behind are pretty similar, just the hoop through utils before the svg markup as a string.

jeremierobert-coveo: Good 👍

PR #18, opened 2019-07-29
Fully support es6

jeremierobert-coveo:
use the package.json module field

### DIFF
--- a/config/webpack.config.js
+++ b/config/webpack.config.js
@@ -6,11 +6,11 @@ module.exports = {
         CoveoJsSearchExtensions: './src/Index'
     },
     output: {
-        path: path.resolve('./bin/js/commonjs'),
+        path: path.resolve('./bin/commonjs'),
         filename: `[name].js`,
         libraryTarget: 'umd',
         library: 'CoveoExtension',
-        publicPath: '/js/commonjs'
+        publicPath: '/commonjs'
     },
     externals: [
         {
--- a/package.json
+++ b/package.json
@@ -4,10 +4,11 @@
     "name": "coveo-search-ui-extensions",
     "version": "0.0.5",
     "description": "Small generic components to extend the functionality of Coveo's Search UI framework.",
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -4,7 +4,8 @@
     "name": "coveo-search-ui-extensions",
     "version": "0.0.5",
     "description": "Small generic components to extend the functionality of Coveo's Search UI framework.",
-    "main": "./bin/js/commonjs/CoveoJsSearchExtensions.js",
+    "main": "./bin/commonjs/CoveoJsSearchExtensions.js",
+    "module": "./bin/es6/index.js",
```
jeremierobert-coveo: That is the important change.

Now it will auto switch to the es6 version when using es6 and the commonjs otherwise.

louis-bompart: Nice! 👍

louis-bompart: Link to doc for ref sake ?

jeremierobert-coveo: https://webpack.js.org/configuration/resolve/#resolvemainfields

C'est aussi support par Rollup.js

louis-bompart: Pas fan du manque d'official npm support but I can live that with that tho

config/webpack.config.js
```diff
@@ -6,11 +6,11 @@ module.exports = {
         CoveoJsSearchExtensions: './src/Index'
     },
     output: {
-        path: path.resolve('./bin/js/commonjs'),
+        path: path.resolve('./bin/commonjs'),
         filename: `[name].js`,
         libraryTarget: 'umd',
         library: 'CoveoExtension',
-        publicPath: '/js/commonjs'
+        publicPath: '/commonjs'
```
mikegron: sa vas modifier le path dans le CDN ca mais jimagine que c'est correct vue que sa vas etre une autre version anyway

jeremierobert-coveo: 👍

PR #20, opened 2019-07-31
Viewed By Customer

louis-bompart:
~~Waiting for coveo/search-ui#1170~~
## Goal

Implement a result component that indicate when a result have been seen by an user (identified when sending the query).

Allow basic customization:
 - Hide/Show the icon
 - Change the label

## Out of scopes, improvement for future PRs
 - Action on click
 - More info on hover.

## Screenshots:

image
image


pages/index.html
```diff
@@ -44,107 +44,93 @@
         </script>
     </head>
 
-    <body id="search" class="CoveoSearchInterface" data-enable-history="true" style="padding: 1em;">
```
erocheleau: Why is the padding gone?

pages/index.html
```diff
@@ -44,107 +44,93 @@
...
-        <span class="CoveoAnalytics"></span>
```
erocheleau: Why is the Analytics gone?

pages/index.html
```diff
@@ -44,107 +44,93 @@
...
-            <div class="coveo-main-section">
-                <div class="coveo-facet-column">
-                    <div class="CoveoResultsFilter" data-text="Show Only Attached Results"></div>
-                    <div class="CoveoFacet" data-title="Type" data-field="@objecttype" data-tab="All"></div>
-                    <div class="CoveoFacet" data-title="FileType" data-field="@filetype" data-tab="All"></div>
-                    <div class="CoveoFacet" data-title="Author" data-field="@author" data-tab="All"></div>
-                    <div class="CoveoFacet" data-title="Year" data-field="@year" data-tab="All"></div>
-                    <div class="CoveoFacet" data-title="Month" data-field="@month" data-tab="All"></div>
+            <div class="coveo-results-column">
+                <div class="CoveoShareQuery"></div>
```
erocheleau: Is this page supposed to mimic a side panel? if so, you could probably remove the CoveoSettings, CoveoShareQuery, CoveoExportToExcel, CoveoPreferencesPanel etc.

If it's a normal page, disregard.

pages/index.html
```diff
@@ -44,107 +44,93 @@
...
-                              </div>
-                              <div class="coveo-result-cell" style="vertical-align:top;text-align:center;width:32px;">
-                                <span class="CoveoIcon" data-small="true" data-with-label="false"></span>
+                    <div class="coveo-sort-section">
+                        <span class="CoveoSort" data-sort-criteria="relevancy" data-caption="Relevance"></span>
+                        <span class="CoveoSort" data-sort-criteria="date descending,date ascending" data-caption="Date"></span>
+                    </div>
+                </div>
+                <div class="CoveoHiddenQuery"></div>
+
```
erocheleau: blank line?

pages/index.html
```diff
@@ -44,107 +44,93 @@
...
+                                <span class="CoveoFieldValue" data-field="@language" data-text-caption="Language" style="margin-right:30px;"></span>
+                                <span class="CoveoFieldValue" data-field="@filetype" data-text-caption="File Type" style="margin-right:30px;"></span>
+                              </div>
+                            </div>
+                            <div class="coveo-result-row" style="margin-top:10px;">
+                              <div class="coveo-result-cell">
+                                <div class="CoveoPrintableUri"></div>
+                              </div>
+                            </div>
+                            <div class="coveo-result-row" style="margin-top:10px;">
```
erocheleau: So by default we put it in a row below each result?

pages/index.html
```diff
@@ -44,107 +44,93 @@
...
+                                <div class="coveo-result-cell">
+                                  <div class="CoveoViewedByCustomer" data-show-icon="true"></div>
```
erocheleau: show icon is true by default, you could remove that attribute there.

src/components/ViewedByCustomer/ViewedByCustomer.ts
```diff
@@ -0,0 +1,65 @@
...
+    };
+
+    /**
+     * Create an instance of {@link ViewedByCustomer}.
+     * @param element Element on which to bind the component.
+     * @param options Initialization options of the component.
+     * @param bindings Bindings of the Search-UI environment.
+     */
+    public constructor(public element: HTMLElement, public options: IViewedByCustomerOptions, public bindings: IComponentBindings) {
+        super(element, ViewedByCustomer.ID, bindings)/* istanbul ignore next Issue with Istanbul and super calls*/;
```
erocheleau: `/* istanbul ignore next Issue with Istanbul and super calls*/`
Is that annotation necessary?
Can we make it a general rule somewhere so we don't have to have annotations in the code?
If we can't can we make the comment line be above the line instead of on the same line?

src/components/ViewedByCustomer/ViewedByCustomer.ts
```diff
@@ -0,0 +1,65 @@
...
+    private render() {
+        if (this.options.showIcon) {
+            const iconElement = document.createElement('span');
+            iconElement.className = 'viewed-by-customer-icon';
+            iconElement.innerHTML = user;
+            $$(this.element).append(iconElement);
+        }
+
+        const labelElement = document.createElement('span');
+        labelElement.className = 'viewed-by-customer-label';
```
erocheleau: Should you be using `labelElement.classList.add(...)` instead to not overwrite any other class that may exist?
Not really necessary but I think I prefer that method to overriding the entire className.
But it's debatable since you create the element yourself.

src/components/ViewedByCustomer/ViewedByCustomer.ts
```diff
@@ -0,0 +1,65 @@
...
+     * Create an instance of {@link ViewedByCustomer}.
+     * @param element Element on which to bind the component.
+     * @param options Initialization options of the component.
+     * @param bindings Bindings of the Search-UI environment.
+     */
+    public constructor(public element: HTMLElement, public options: IViewedByCustomerOptions, public bindings: IComponentBindings) {
+        super(element, ViewedByCustomer.ID, bindings)/* istanbul ignore next Issue with Istanbul and super calls*/;
+        this.options = ComponentOptions.initComponentOptions(element, ViewedByCustomer, options);
+
+        if (this.resolveResult().isUserActionView) {
```
erocheleau: Where is this `isUserActionView` coming from? 
Is it from the generic "Component" from the jsui?

I remember we talked at some point about potentially greying out the result when it had been already viewed, any opinion on adding a potential CSS class to the entire result rendered DOM when it is visited?

erocheleau: is resolveResult safe?
Will this assert if the result is null for example or something like that?

tests/components/ViewedByCustomer/ViewedByCustomer.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+            expect(testComponent.element.getElementsByClassName('viewed-by-customer-icon').length).toBe(1);
+            expect(testComponent.element.getElementsByClassName('viewed-by-customer-label').length).toBe(1);
+            expect(testComponent.element.getElementsByClassName('viewed-by-customer-label').item(0).innerHTML).toBe('Viewed by Customer');
+        });
+
+        it('should display the label given in the option if any is given', () => {
+            const option = new Mock.AdvancedComponentSetupOptions();
+            (option.cmpOptions as IViewedByCustomerOptions) = { label: 'test label' };
+            const testComponent = Mock.advancedResultComponentSetup<ViewedByCustomer>(ViewedByCustomer, fakeResult, option).cmp;
+            expect(testComponent.element.getElementsByClassName('viewed-by-customer-label').length).toBe(1);
```
erocheleau: constants for this please `'viewed-by-customer-label'`

tests/components/ViewedByCustomer/ViewedByCustomer.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+import { IViewedByCustomerOptions } from '../../../src/components/ViewedByCustomer/ViewedByCustomer';
+import { IQueryResult } from 'coveo-search-ui';
+
+describe('ViewedByCustomer', () => {
+    describe('when the field isUserActionView of the result true', () => {
+        const fakeResult: IQueryResult = { ...Fake.createFakeResult(), isUserActionView: true };
+
+        it('should display the default label and the icon when no options are given', () => {
+            const testComponent = Mock.advancedResultComponentSetup<ViewedByCustomer>(ViewedByCustomer, fakeResult).cmp;
+            expect(testComponent.element.getElementsByClassName('viewed-by-customer-icon').length).toBe(1);
```
erocheleau: Constant for this please `'viewed-by-customer-icon'`

tests/components/ViewedByCustomer/ViewedByCustomer.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+            (option.cmpOptions as IViewedByCustomerOptions) = { showIcon: false };
+            const testComponent = Mock.advancedResultComponentSetup<ViewedByCustomer>(ViewedByCustomer, fakeResult, option).cmp;
+            expect(testComponent.element.getElementsByClassName('viewed-by-customer-icon').length).toBe(0);
+        });
+    });
+
+    describe('when the field isUserActionView of the result is false', () => {
+        const fakeResult: IQueryResult = { ...Fake.createFakeResult(), isUserActionView: false };
+
+        it('should hide the icon if showIcon option is set to false', () => {
```
erocheleau: Wait... it should hide the icon EVEN when the showIcon is true if the `isUserActionView` is false right?

package.json
```diff
@@ -48,7 +48,7 @@
         "@types/jasmine": "^3.3.9",
         "@types/node": "^11.11.3",
         "@types/sinon": "^7.0.13",
-        "coveo-search-ui": "^2.6459.0",
+        "coveo-search-ui": "2.0.342",
```
jeremierobert-coveo: Why do we rollback version ?

src/components/ViewedByCustomer/ViewedByCustomer.ts
```diff
@@ -0,0 +1,65 @@
+import { Component, ComponentOptions, IComponentBindings, $$, Initialization } from 'coveo-search-ui';
+import { user } from '../../utils/icons';
+
+/**
+ * The options for the ViewedByCustomerComponent
+ */
+export interface IViewedByCustomerOptions {
+    showIcon?: boolean;
```
jeremierobert-coveo: Juicy comments on each options

src/components/ViewedByCustomer/ViewedByCustomer.ts
```diff
@@ -0,0 +1,65 @@
...
+            this.render();
+        }
+    }
+
+    private render() {
+        if (this.options.showIcon) {
+            const iconElement = document.createElement('span');
+            iconElement.className = 'viewed-by-customer-icon';
+            iconElement.innerHTML = user;
+            $$(this.element).append(iconElement);
```
jeremierobert-coveo: Why $$ for an append ?

`this.element.appendChild(iconElement);`

src/components/ViewedByCustomer/ViewedByCustomer.ts
```diff
@@ -0,0 +1,65 @@
...
+        }
+
+        const labelElement = document.createElement('span');
+        labelElement.className = 'viewed-by-customer-label';
+        labelElement.innerText = this.options.label;
+        $$(this.element).append(labelElement);
```
jeremierobert-coveo: Same as Above

src/components/ViewedByCustomer/ViewedByCustomer.ts
jeremierobert-coveo: `labelElement.classList.add` seams a better option

src/components/ViewedByCustomer/ViewedByCustomer.ts
```diff
@@ -0,0 +1,65 @@
...
+
+        if (this.resolveResult().isUserActionView) {
+            this.render();
+        }
+    }
+
+    private render() {
+        if (this.options.showIcon) {
+            const iconElement = document.createElement('span');
+            iconElement.className = 'viewed-by-customer-icon';
```
jeremierobert-coveo: `.classList.add`

jeremierobert-coveo: const of static for this magic string

tests/components/ViewedByCustomer/ViewedByCustomer.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+
+describe('ViewedByCustomer', () => {
+    describe('when the field isUserActionView of the result true', () => {
+        const fakeResult: IQueryResult = { ...Fake.createFakeResult(), isUserActionView: true };
+
+        it('should display the default label and the icon when no options are given', () => {
+            const testComponent = Mock.advancedResultComponentSetup<ViewedByCustomer>(ViewedByCustomer, fakeResult).cmp;
+            expect(testComponent.element.getElementsByClassName('viewed-by-customer-icon').length).toBe(1);
+            expect(testComponent.element.getElementsByClassName('viewed-by-customer-label').length).toBe(1);
+            expect(testComponent.element.getElementsByClassName('viewed-by-customer-label').item(0).innerHTML).toBe('Viewed by Customer');
```
jeremierobert-coveo: `.innerText` plz you dont validate the HTML but the TEXT

tests/components/ViewedByCustomer/ViewedByCustomer.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+        });
+
+        it('should display the label given in the option if any is given', () => {
+            const option = new Mock.AdvancedComponentSetupOptions();
+            (option.cmpOptions as IViewedByCustomerOptions) = { label: 'test label' };
```
jeremierobert-coveo: const for you test label and also randomize it.

tests/components/ViewedByCustomer/ViewedByCustomer.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+            const testComponent = Mock.advancedResultComponentSetup<ViewedByCustomer>(ViewedByCustomer, fakeResult, option).cmp;
+            expect(testComponent.element.getElementsByClassName('viewed-by-customer-label').length).toBe(1);
+            expect(testComponent.element.getElementsByClassName('viewed-by-customer-label').item(0).innerHTML).toBe('test label');
```
jeremierobert-coveo: `innerText` plz

tests/components/ViewedByCustomer/ViewedByCustomer.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+        });
+
+        it('should hide the icon if showIcon option is set to false', () => {
+            const option = new Mock.AdvancedComponentSetupOptions();
+            (option.cmpOptions as IViewedByCustomerOptions) = { showIcon: false };
```
jeremierobert-coveo: not sure about the cast, can't you use the constructor instead ?

tests/components/ViewedByCustomer/ViewedByCustomer.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+            expect(testComponent.element.getElementsByClassName('viewed-by-customer-icon').length).toBe(0);
+        });
+    });
+
+    describe('when the field isUserActionView of the result is false', () => {
+        const fakeResult: IQueryResult = { ...Fake.createFakeResult(), isUserActionView: false };
+
+        it('should hide the icon if showIcon option is set to false', () => {
+            const option = new Mock.AdvancedComponentSetupOptions();
+            (option.cmpOptions as IViewedByCustomerOptions) = { showIcon: false };
```
jeremierobert-coveo: Same as line 27

tests/components/ViewedByCustomer/ViewedByCustomer.spec.ts
```diff
@@ -0,0 +1,43 @@
+import { ViewedByCustomer } from '../../../src/Index';
```
jeremierobert-coveo: What happen if **isUserActionView** is not present on the result ?

pages/index.html
louis-bompart: It's a normal page, based on the JSUI one. The only changes that I've done there is to remove duplicated main-section (This comment apply mostly to every of you comment on this file I think)

src/components/ViewedByCustomer/ViewedByCustomer.ts
louis-bompart: It's from `IQueryResult` of the coveo/search-ui. See coveo/search-ui#1170

louis-bompart: There's an assert :

[Component.ts](https://github.com/coveo/search-ui/blob/master/src/ui/Base/Component.ts#L363)
```typescript
  public resolveResult(): IQueryResult {
    return Component.getResult(this.element);
  }
```
[Component.ts](https://github.com/coveo/search-ui/blob/master/src/ui/Base/Component.ts#L408)
```typescript
  static getResult(element: HTMLElement, noThrow: boolean = false): IQueryResult {
    var resultElement = $$(element).closest('.CoveoResult');
    Assert.check(noThrow || resultElement != undefined);
    return resultElement['CoveoResult'];
  }```

We look up for the first result by ascending the DOM tree from the Viewed By Customer element, if we do not found a result, being a component that should be used on a ResultTemplate, that's a legitimate usage issue and the Assert will throw rightfully so.

File: tests/components/ViewedByCustomer/ViewedByCustomer.spec.ts (hunk 12 above)
Comment (louis-bompart) [reply tests/components/ViewedByCustomer/ViewedByCustomer.spec.ts]: Absolutely, I'll check that out 🤦‍♂

File: package.json (hunk 13 above)
Comment (louis-bompart) [reply package.json]: JSUI beta package is on 2.0.x

File: src/components/ViewedByCustomer/ViewedByCustomer.ts (hunk 7 above)
Comment (louis-bompart) [reply src/components/ViewedByCustomer/ViewedByCustomer.ts]: No we cannot, it's an issue with Istanbul that consider the semi at the end of the super as a statement (it baffles me as much as you). So, istanbul ignore next ignoring the next statement does not help if we don't put it just before the 'statement', so, here, just before the semicolon.

File: src/components/ViewedByCustomer/ViewedByCustomer.ts (hunk 14 above)
Comment (jeremierobert-coveo) [reply src/components/ViewedByCustomer/ViewedByCustomer.ts]: Do it in the type so intellisense can catch it

File: tests/components/ViewedByCustomer/ViewedByCustomer.spec.ts (hunk 19 above)
Comment (louis-bompart) [reply tests/components/ViewedByCustomer/ViewedByCustomer.spec.ts]: I disagree: randomizing the label does not add value to the test and reduce the readability.
Same for the use of a const.

In my opinion:
The phases of a test (setup, exercise, verify, teardown) should be as loosely connected to each other, with the notable exception of setup and teardown, because they can sometimes be grouped up (3-phase test v.s. 4-phase test, same fight to be fair). In the present case, l19 is part of the exercise phase: We create the component following the instruction of the test (we put a custom label in the options). l22 is in the verify phase: We check that the label is the good one. If they were some treatment on the string, we would not want to reproduce it in the test, we'd want to see in just a quick sight what's going in, what's going out.

File: tests/components/ViewedByCustomer/ViewedByCustomer.spec.ts (hunk 21 above)
Comment (louis-bompart) [reply tests/components/ViewedByCustomer/ViewedByCustomer.spec.ts]: To be fair, the cast is mostly for typing, it does not need to exists.
Why not set directly in the constructor: because I can't only set the option without setting the element because cmpOptions is the second attribute.
https://github.com/coveo/search-ui-tests/blob/master/src/MockEnvironment/AdvancedComponentsSetup.ts#L44

Comment (jeremierobert-coveo) [reply tests/components/ViewedByCustomer/ViewedByCustomer.spec.ts]: yes you can `new Mock.AdvancedComponentSetupOptions(null, { showIcon: false });`

File: tests/components/ViewedByCustomer/ViewedByCustomer.spec.ts (hunk 23 above)
Comment (louis-bompart) [reply tests/components/ViewedByCustomer/ViewedByCustomer.spec.ts]: Will add section for 'falsy' values.

File: src/components/ViewedByCustomer/ViewedByCustomer.ts [hunk 24]
```diff
@@ -0,0 +1,65 @@
...
+     */
+    public static readonly options: IViewedByCustomerOptions = {
+        /**
+         * If true, will display an icon when the component is displayed itself.
+         */
+        showIcon: ComponentOptions.buildBooleanOption({ defaultValue: true }),
+        /**
+         * The label that should be displayed when the component is displayed.
+         */
+        label: ComponentOptions.buildStringOption({ defaultValue: 'Viewed by Customer' })
```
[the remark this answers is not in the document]
jeremierobert-coveo: option + localization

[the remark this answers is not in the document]
jeremierobert-coveo: Genre `l(this.option.label)`

src/components/ViewedByCustomer/ViewedByCustomer.ts
louis-bompart: One or the other, not both, tho: DRY.

src/components/ViewedByCustomer/ViewedByCustomer.ts
erocheleau: So what about the answer to my other question?

src/components/ViewedByCustomer/ViewedByCustomer.ts
erocheleau: I would also prefer not relying on $$ if you don't have to.

tests/components/ViewedByCustomer/ViewedByCustomer.spec.ts
erocheleau: I agree, don't randomize the value, but make sure it's not the default and that's enough. Randomizing a value adds nothing.

If you want to be really in the details, try with "dirty" strings to make sure it works.

Strings with É &; stuff <script> if you want to :)

src/components/ViewedByCustomer/ViewedByCustomer.ts
louis-bompart: About your other question, as you said this was talked about but never specified (in the 'specification' meaning), thus, not implemented there.

More importantly, this is the bare-MVP for ViewedByCustomer, I have no objection to improve it later, but, well, _later_.

pages/index.html
louis-bompart: Btw, just to clarify: `pages/index.html` is used for development only.

src/components/ViewedByCustomer/ViewedByCustomer.scss
```diff
@@ -0,0 +1,18 @@
+@import '../../sass/Variables.scss';
+@import '../../sass/Text.scss';
+
+.CoveoViewedByCustomer {
+    color: black;
+    font-weight: bold;
+    font-size: 14px;
```
jeremierobert-coveo: why 14px ?
Why not 1em ?

louis-bompart: We don't use em to fix font size in the search-ui & co.

src/components/ViewedByCustomer/ViewedByCustomer.ts
```diff
@@ -0,0 +1,71 @@
...
+    /**
+     * Unique Identifier used by the Search-UI.
+     */
+    public static readonly ID = 'ViewedByCustomer';
+
+    /**
+     * Default options used by the component.
+     */
+    public static readonly options: IViewedByCustomerOptions = {
+        showIcon: ComponentOptions.buildBooleanOption({ defaultValue: true }),
```
flapierre-coveo: I know you just moved the comments to the interface, but if I remember right, the comments need to be on the static options for the interface editor to display them.

tests/components/ViewedByCustomer/ViewedByCustomer.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+import { Mock, Fake } from 'coveo-search-ui-tests';
+import { IViewedByCustomerOptions } from '../../../src/components/ViewedByCustomer/ViewedByCustomer';
+import { IQueryResult } from 'coveo-search-ui';
+import { AdvancedComponentSetupOptions } from 'coveo-search-ui-tests/MockEnvironment';
+
+describe('ViewedByCustomer', () => {
+    const VIEWED_BY_CUSTOMER_LABEL_CSS_CLASS = 'viewed-by-customer-label';
```
flapierre-coveo: Is it necessary to prefix with VIEWED_BY_CUSTOMER? I feel like, since we are in the context of a test file for ViewedByCustomer, it makes the asserts too verbose

tests/components/ViewedByCustomer/ViewedByCustomer.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+            const testComponent = Mock.advancedResultComponentSetup<ViewedByCustomer>(ViewedByCustomer, fakeResult, option).cmp;
+            expect(testComponent.element.getElementsByClassName(VIEWED_BY_CUSTOMER_ICON_CSS_CLASS).length).toBe(0);
+        });
+    });
+
+    [false, undefined, null].forEach(falsyValue => {
+        describe(`when the field isUserActionView of the result is ${falsyValue}`, () => {
+            const fakeResult: IQueryResult = { ...Fake.createFakeResult(), isUserActionView: falsyValue };
+
+            it('nothing should should be displayed or added to the DOM', () => {
```
flapierre-coveo: Rephrase the test to start with `it should...` rather than `it nothing...` something like `it should not display the component`

src/components/ViewedByCustomer/ViewedByCustomer.ts
```diff
@@ -0,0 +1,71 @@
...
+        label: ComponentOptions.buildStringOption({ defaultValue: l(`${ViewedByCustomer.ID}_DefaultLabel`) })
+    };
+
+    //#region Internal CSS selectors
```
flapierre-coveo: Not sure if it's worth putting #region for 2 lines of code

src/components/ViewedByCustomer/ViewedByCustomer.ts
louis-bompart: The interface editor is not supported (Unless they have been massive simplification) in this PR, so I do not think we should bother with this question right away, because we'll be unable to verify.

pages/index.html
erocheleau: You didn't answer...

pages/index.html
erocheleau: It can still be a model for what kind of page will need to be created no?

erocheleau: In the jsui, the folder `pages/` contains a lot of example pages that people sometimes copy and paste.

pages/index.html
louis-bompart: As I said, it's based on the Search page template of the coveo/search-ui.
https://github.com/coveo/search-ui/blob/master/pages/All.html I invite to check it out, you won't find an Analytics.

pages/index.html
louis-bompart: I gave you a general answer to your question about your file in a previous comment:
This file is only used for testing purposes. I based my template on the one from the JSUI and cleaned it because there was a main-section in another main section + duplicate code and all.

I don't think there's real value to be gained to further modify it because if you need it for development you're bound to customize it further.

Also my previous comment apply + the one of Analytic:
> As I said, it's based on the Search page template of the coveo/search-ui.
https://github.com/coveo/search-ui/blob/master/pages/All.html I invite to check it out, you won't find an Analytics.

PR #21, opened 2019-08-01
ES6 dual package requirements

jeremierobert-coveo:
Just do some modifications to fix nitty-gritty issues I found.

And add some padding to the user action no action message.


scripts/before.deploy.js
```diff
@@ -1,12 +1,12 @@
 const { readFileSync, writeFileSync } = require('fs');
 const { resolve } = require('path');
 
-const BRANCH = process.env.TRAVIS_BRANCH;
-if (!BRANCH) {
-    throw new Error('TRAVIS_BRANCH is not define in the current environment');
+const TAG = process.env.TRAVIS_TAG;
+if (!TAG) {
+    throw new Error('TRAVIS_TAG is not define in the current environment');
```
louis-bompart: defined :P

src/components/UserActions/UserActions.scss
```diff
@@ -44,6 +44,11 @@ $search-ui-dropdown-header-color: #1d4f76;
         margin: 0.5em 0;
     }
 
+    .coveo-no-actions {
```
louis-bompart: Euh? Out of scope I'd say ?

jeremierobert-coveo: Can we include this in the scope ?

louis-bompart: We can, sure, but can you update your pr description accordingly etc

PR #22, opened 2019-08-05
Change the way UserProfileModel is initialize

jeremierobert-coveo:
I change the way to initilize the model of the user action as an atempt to simplify the process.


src/components/UserActions/ClickedDocumentList.ts
```diff
@@ -90,13 +90,18 @@ export class ClickedDocumentList extends Component {
         super(element, ClickedDocumentList.ID, bindings);
```
louis-bompart:
```suggestion
        super(element, ClickedDocumentList.ID, bindings) /* istanbul ignore next Istanbul issue with next*/;
```

Just so we got 💯 coverage (I know we don't need nor want to aim 💯 but we can, however, aim to that and improve the coverage by removing some erratic thing like this issue with Istanbul.

src/models/UserProfileModel.ts
```diff
@@ -62,20 +69,18 @@ export class UserProfileModel extends Model {
...
+        Assert.isNotUndefined(this.options.organizationId);
+        Assert.isNotUndefined(this.options.accessToken);
 
         this.getOrFetchCache = {};
 
         this.endpoint = new UserProfilingEndpoint({
             uri: this.options.restUri,
-            accessToken: SEARCH_INTERFACE_ENDPOINT.accessToken,
-            organization: this.options.organization
+            accessToken: this.options.accessToken || (this.options.searchEndpoint as SearchEndpoint).accessToken,
```
louis-bompart: Why the `as SearchEndpoint`, is accessToken not a prop of `ISearchEndpoint` ?

tests/utils.ts
```diff
@@ -79,3 +61,8 @@ export function delay(callback: () => any, timemout = 0) {
         }, timemout);
     });
 }
+
+export function fakeUserProfileModel(root: HTMLElement, sandbox: SinonSandbox) {
```
louis-bompart: Juicy comment for this ;)

src/models/UserProfileModel.ts
jeremierobert-coveo: Exactly only the class as accessToken

src/components/UserActions/ClickedDocumentList.ts
jeremierobert-coveo: k

src/models/UserProfileModel.ts
louis-bompart: Next question, this.options.searchEndpoint is always a defined object?

jeremierobert-coveo: Yes always

PR #25, opened 2019-08-14
 Only mark manual query as User Query SFINT-2449

jeremierobert-coveo:
Only search event that is defined as manual will be tag as a User Query instead of each search event that as a query_expression.


src/utils/events.ts
```diff
@@ -0,0 +1,8 @@
+export const MANUAL_SEARCH_EVENT_CAUSE = Object.freeze([
+    'omniboxAnalytics',
```
jeremierobert-coveo: Here, I not 100% sure which event cause shall be mark as a User Query. I mean for submit is trivial but others...

louis-bompart:
> See SFINT-2449

Yeah nah, not enough, details.

src/utils/events.ts
jeremierobert-coveo: When I say not 100% sure mean, I'm open to discussion 😄 .
I passed each listed search cause in the documentation and those are the ones that came out.

louis-bompart: I have no real opinions on the matter to be perfectly honest, just'd like to have a good confidence in it :)

src/components/UserActions/Strings.ts
```diff
@@ -17,8 +17,8 @@ Translation.register(Language.English, {
     UserActivity_other_event: 'Other Event',
     UserActivity_other_events: 'Other Events',
 
-    UserActivity_search: 'User Query',
-    UserActivity_query: 'Query',
+    UserActivity_search: 'Query',
+    UserActivity_query: 'User Query',
```
tedre191: What's changed exactly ?

src/components/UserActions/UserActivity.ts
```diff
@@ -349,6 +350,10 @@ export class UserActivity extends Component {
 
         return cell;
     }
+
+    private isManualSubmitAction(action: UserAction) {
```
tedre191: Add a comment that explain what type of events are "manual"

tests/components/UserActions/UserActivity.spec.ts
```diff
@@ -201,20 +201,33 @@ describe('UserActivity', () => {
...
-                    return env;
-                })
-            );
-
-            return delay(() => {
-                const clickElement = mock.cmp.element.querySelector('.coveo-search');
-
-                expect(clickElement).not.toBeNull();
-                expect(clickElement.querySelector<HTMLElement>('.coveo-title').innerText).toBe('User Query');
+        ['omniboxAnalytics', 'userActionsSubmit', 'omniboxFromLink', 'searchboxAsYouType', 'searchboxSubmit', 'searchFromLink'].map(cause => {
```
tedre191: @louis-bompart is gonna love that, but I don't mind.

louis-bompart: I'm the one that asked for it :D

src/components/UserActions/Strings.ts
jeremierobert-coveo: The label tagged in the UserActivity component.

PR #26, opened 2019-08-14
Disable User Actions when userId is falsy SFINT-2453

jeremierobert-coveo:
See SFINT-2453

Fix an issue where an empty string can be use as a userId

### DIFF
--- a/src/components/UserActions/UserActions.ts
+++ b/src/components/UserActions/UserActions.ts
@@ -81,16 +81,20 @@ export class UserActions extends Component {
 
         this.options = ComponentOptions.initComponentOptions(element, UserActions, options);
 
-        (get(this.root, UserProfileModel) as UserProfileModel)
-            .getActions(this.options.userId)
-            .then(actions => (actions.length > 0 ? this.render() : this.renderNoActions()))
-            .catch(() => this.renderNoActions());
+        if (!this.options.userId) {
+            this.disable();
+        } else {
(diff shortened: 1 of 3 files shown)


src/components/UserActions/UserActions.ts
```diff
@@ -81,16 +81,20 @@ export class UserActions extends Component {
```
jeremierobert-coveo: Side by side diff is advise

src/components/UserActions/UserActions.ts
```diff
@@ -81,16 +81,20 @@ export class UserActions extends Component {
 
         this.options = ComponentOptions.initComponentOptions(element, UserActions, options);
 
-        (get(this.root, UserProfileModel) as UserProfileModel)
-            .getActions(this.options.userId)
-            .then(actions => (actions.length > 0 ? this.render() : this.renderNoActions()))
-            .catch(() => this.renderNoActions());
+        if (!this.options.userId) {
+            this.disable();
```
louis-bompart: Because you're using a precondition you can remove the else scope if you add a return; after the `this.disable()`.

This will streamline the code even more imo.

jeremierobert-coveo: Even in a constructor ?

louis-bompart: I don't see why not, if it's an anonymous return.

PR #27, opened 2019-08-15
Use string locales SFINT-2448

jeremierobert-coveo:
This prevent an issue in the interface editor where the registered strings are nuke by an other definition of `toLocaleString` loaded. (Multiple CoveoJsSearch.js loaded)


jeremierobert-coveo:
We define the stuff if it's not define so can test it but it does not bring anything.

louis-bompart:
> We define the stuff if it's not define so can test it but it does not bring anything.

I agree, so we can ignore the other cases

PR #28, opened 2019-08-15
new Events and localization for AttachResult and ResultsFilter components

mikegron:
Advise if I should add all events/strings in a single file for every component instead ?

Added strings for defaults strings of AttachResult and ResultsFilter
Added events when attaching / detaching result and filtering
SNOW-182

### DIFF
--- a/src/components/AttachResult/AttachResult.ts
+++ b/src/components/AttachResult/AttachResult.ts
@@ -8,9 +8,12 @@ import {
     IAnalyticsCaseAttachMeta,
     IAnalyticsCaseDetachMeta,
     analyticsActionCauseList,
-    IAnalyticsActionCause
+    IAnalyticsActionCause,
+    l
 } from 'coveo-search-ui';
 import { paperclipIcon } from '../../utils/icons';
+import { AttachResultEvents, IAttachResultEventArgs } from './Events';
+import './Strings';
 
 export interface IAttachResultOptions {
     /** Specifies the tooltip displayed when the result is not attached. */
@@ -37,12 +40,12 @@ export class AttachResult extends Component {
     private buttonElement: HTMLElement;
     private tooltipElement: HTMLElement;
 
-    static options: IAttachResultOptions = {
+    public static readonly options: IAttachResultOptions = {
         attachCaption: ComponentOptions.buildStringOption({
-            defaultValue: 'Attach Result'
+            defaultValue: l(`${AttachResult.ID}_DefaultAttachCaption`)
         }),
         detachCaption: ComponentOptions.buildStringOption({
-            defaultValue: 'Detach Result'
+            defaultValue: l(`${AttachResult.ID}_DefaultDetachCaption`)
         }),
         attach: ComponentOptions.buildCustomOption(
             name => (result: IQueryResult) =>
@@ -111,6 +114,7 @@ export class AttachResult extends Component {
             .then(() => {
                 this.attached = true;
                 this.logAnalyticsCaseEvent(analyticsActionCauseList.caseAttach);
+                Coveo.$$(this.root).trigger(AttachResultEvents.Attach, { queryResult: this.queryResult } as IAttachResultEventArgs);
             })
             .finally(() => {
                 this.setLoading(false);
@@ -131,6 +135,7 @@ export class AttachResult extends Component {
             .then(() => {
                 this.attached = false;
                 this.logAnalyticsCaseEvent(analyticsActionCauseList.caseDetach);
+                Coveo.$$(this.root).trigger(AttachResultEvents.Detach, { queryResult: this.queryResult } as IAttachResultEventArgs);
             })
             .finally(() => {
                 this.setLoading(false);
--- a/src/components/AttachResult/AttachResult.ts
+++ b/src/components/AttachResult/AttachResult.ts
@@ -42,10 +42,10 @@ export class AttachResult extends Component {
 
     public static readonly options: IAttachResultOptions = {
         attachCaption: ComponentOptions.buildStringOption({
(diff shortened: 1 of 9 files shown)


louis-bompart:
Take the habit of creating branches from your fork :P

tests/components/ResultsFilter/ResultsFilter.spec.ts
```diff
@@ -42,6 +43,23 @@ describe('ResultsFilter', () => {
         expect(filter.cmp.isSelected()).toBeFalsy();
     });
 
+    it('should trigger events on toggling', done => {
+        let cpt = 0;
+        Coveo.$$(filter.env.root).on(ResultsFilterEvents.Click, (evt: Event, args: IResultsFilterEventArgs) => {
+            expect(evt.type).toBe(ResultsFilterEvents.Click);
+            if (args.checked) {
```
louis-bompart: What's the point of this if-else ?

tests/components/ResultsFilter/ResultsFilter.spec.ts
```diff
@@ -42,6 +43,23 @@ describe('ResultsFilter', () => {
...
+                cpt++;
+            } else {
+                cpt++;
+            }
+            if (cpt == 2) {
+                done();
+            }
+        });
+        filter.cmp.toggle();
```
louis-bompart: Double toggle is normal?

src/components/ResultsFilter/Strings.ts
```diff
@@ -0,0 +1,5 @@
+import { Translation, Language } from '../../utils/translation';
```
louis-bompart: Need to add this file to the side-effects of package.json

src/components/AttachResult/Strings.ts
```diff
@@ -0,0 +1,6 @@
+import { Translation, Language } from '../../utils/translation';
```
louis-bompart: Need to add this file to the side-effects of package.json

tests/components/ResultsFilter/ResultsFilter.spec.ts
mikegron: Make sure when I toggle two times the event has both values

tests/components/ResultsFilter/ResultsFilter.spec.ts
mikegron: yeah alltough I could split this test in two maybe

src/components/AttachResult/Strings.ts
mikegron: what does it do exactly ? 🤔

jeremierobert-coveo: That's false

tests/components/ResultsFilter/ResultsFilter.spec.ts
louis-bompart: Yeah but... You have exactly the same statements in the if and else

tests/components/ResultsFilter/ResultsFilter.spec.ts
```diff
@@ -42,6 +43,23 @@ describe('ResultsFilter', () => {
         expect(filter.cmp.isSelected()).toBeFalsy();
     });
 
+    it('should trigger events on toggling', done => {
+        let cpt = 0;
+        Coveo.$$(filter.env.root).on(ResultsFilterEvents.Click, (evt: Event, args: IResultsFilterEventArgs) => {
```
jeremierobert-coveo: I think you can simplify this by giving a spy as callback ?

src/components/AttachResult/Strings.ts
louis-bompart: To be fairly honest, I'm quite new to the concept of tree shaking and side-effects, so I won't be the best to explains, but I can give you a link to the documentation: https://webpack.js.org/guides/tree-shaking/ 😅

tests/components/ResultsFilter/ResultsFilter.spec.ts
```diff
@@ -42,6 +44,33 @@ describe('ResultsFilter', () => {
...
+            });
+            filter.cmp.toggle();
+        });
+    });
+
+    it('should trigger event with checked false when toggling twice', () => {
+        return new Promise(resolve => {
+            filter.cmp.toggle();
+            Coveo.$$(filter.env.root).on(ResultsFilterEvents.Click, (evt: Event, args: IResultsFilterEventArgs) => {
+                expect(evt.type).toBe(ResultsFilterEvents.Click);
```
louis-bompart: Not sure if there's real value on this expect.

tests/components/ResultsFilter/ResultsFilter.spec.ts
```diff
@@ -42,6 +44,33 @@ describe('ResultsFilter', () => {
         expect(filter.cmp.isSelected()).toBeFalsy();
     });
 
+    it('should trigger event with checked true when first toggling', () => {
+        return new Promise(resolve => {
+            Coveo.$$(filter.env.root).on(ResultsFilterEvents.Click, (evt: Event, args: IResultsFilterEventArgs) => {
+                expect(evt.type).toBe(ResultsFilterEvents.Click);
```
louis-bompart: Not sure if there's real value on this expect.

src/components/AttachResult/AttachResult.ts
```diff
@@ -37,12 +40,12 @@ export class AttachResult extends Component {
     private buttonElement: HTMLElement;
     private tooltipElement: HTMLElement;
 
-    static options: IAttachResultOptions = {
+    public static readonly options: IAttachResultOptions = {
```
louis-bompart: Add a JSDOC comment here

PR #29, opened 2019-08-19
Remove duplicate Queries and duplicate Clicked Documents SFINT-2464

jeremierobert-coveo:
Remove duplicate Queries and duplicate Clicked Documents.
Also preserve order

### DIFF
--- a/src/components/UserActions/ClickedDocumentList.ts
+++ b/src/components/UserActions/ClickedDocumentList.ts
@@ -10,7 +10,7 @@ import {
     l,
     get
 } from 'coveo-search-ui';
-import { UserProfileModel } from '../../models/UserProfileModel';
+import { UserProfileModel, UserAction } from '../../models/UserProfileModel';
 import { ExpandableList } from './ExpandableList';
 import { UserActionType } from '../../rest/UserProfilingEndpoint';
 import './Strings';
@@ -98,13 +98,17 @@ export class ClickedDocumentList extends Component {
                 .filter(action => action.document && action.type === UserActionType.Click)
                 .sort((a, b) => a.timestamp.getTime() - b.timestamp.getTime())
                 .reverse()
-                .reduce((acc, action) => (!acc.find(existing => existing.raw.uri_hash === action.raw.uri_hash) ? [...acc, action] : acc), [])
+                .reduce(this.filterDuplicatesClickAction, [])
                 .map(action => {
                     action.document.searchInterface = this.searchInterface;
                     return action.document;
                 });
             this.render();
         }, this.logger.error.bind(this.logger));
+    }
+
+    private filterDuplicatesClickAction(accumulator: UserAction[], action: UserAction): UserAction[] {
+        return !accumulator.find(existing => existing.raw.uri_hash === action.raw.uri_hash) ? [...accumulator, action] : accumulator;
     }
 
     private render() {
--- a/src/components/UserActions/ClickedDocumentList.ts
+++ b/src/components/UserActions/ClickedDocumentList.ts
@@ -98,6 +98,7 @@ export class ClickedDocumentList extends Component {
                 .filter(action => action.document && action.type === UserActionType.Click)
                 .sort((a, b) => a.timestamp.getTime() - b.timestamp.getTime())
                 .reverse()
+                .reduce((acc, action) => (!acc.find(existing => existing.raw.uri_hash === action.raw.uri_hash) ? [...acc, action] : acc), [])
                 .map(action => {
                     action.document.searchInterface = this.searchInterface;
                     return action.document;
--- a/src/components/UserActions/QueryList.ts
+++ b/src/components/UserActions/QueryList.ts
@@ -97,7 +97,8 @@ export class QueryList extends Component {
                 .filter(action => action.query)
                 .sort((a, b) => a.timestamp.getTime() - b.timestamp.getTime())
                 .reverse()
-                .map(action => action.query);
+                .map(action => action.query)
+                .reduce((acc, query) => (acc.indexOf(query) === -1 ? [...acc, query] : acc), []);
             this.render();
         }, this.logger.error.bind(this.logger));
     }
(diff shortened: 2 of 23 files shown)


src/components/UserActions/QueryList.ts
```diff
@@ -88,7 +88,7 @@ export class QueryList extends Component {
      * @param bindings Bindings of the Search-UI environment.
      */
     constructor(public element: HTMLElement, public options: IQueryListOptions, public bindings: IComponentBindings) {
-        super(element, QueryList.ID, bindings);
+        super(element, QueryList.ID, bindings) /* istanbul ignore next Istanbul issue with next */;
```
tedre191: why is istanbul not happy here ?

src/components/UserActions/ClickedDocumentList.ts
```diff
@@ -98,6 +98,7 @@ export class ClickedDocumentList extends Component {
                 .filter(action => action.document && action.type === UserActionType.Click)
                 .sort((a, b) => a.timestamp.getTime() - b.timestamp.getTime())
                 .reverse()
+                .reduce((acc, action) => (!acc.find(existing => existing.raw.uri_hash === action.raw.uri_hash) ? [...acc, action] : acc), [])
```
tedre191: I would add a comment and even split that into a seperate function
ex. `.reduce(filterDuplicates)` or something

src/components/UserActions/QueryList.ts
jeremierobert-coveo: istanbul can't digest supper call.

louis-bompart:
PR should be well presented (i.e. a fitting name and a description if needed)

tests/components/UserActions/ClickedDocumentList.spec.ts
```diff
@@ -9,6 +9,7 @@ import { UserActionType } from '../../../src/rest/UserProfilingEndpoint';
 describe('ClickedDocumentList', () => {
     const TEST_CLICKS = generate(20, i => {
         const document = Fake.createFakeResult();
+        document.uri = `document${Math.random()}`;
```
louis-bompart: We talked about it in a former review with @erocheleau: what's the point of Random in tests. That's, imho, undesirable and useless complexity.

tests/components/UserActions/ClickedDocumentList.spec.ts
```diff
@@ -117,6 +118,54 @@ describe('ClickedDocumentList', () => {
         });
     });
 
+    it('should not show the same query twice', () => {
```
louis-bompart: This test is quite long, maybe put comment to indicate where are the phases of the tests and explain what you're doing.

tests/components/UserActions/ClickedDocumentList.spec.ts
```diff
@@ -117,6 +118,54 @@ describe('ClickedDocumentList', () => {
...
+            if (button) {
+                button.click();
+            }
+
+            const list = mock.env.element.querySelector<HTMLOListElement>('.coveo-list');
+            expect(list.childElementCount).toBe(new Set(TEST_DUPLICATE.map(action => action.raw.uri_hash)).size);
+
+            // Check that the order is respected.
+            const sortedActions = TEST_DUPLICATE.reverse().reduce(
+                (acc, action) => (!acc.find(existing => existing.raw.uri_hash === action.raw.uri_hash) ? [...acc, action] : acc),
```
louis-bompart: No, you're reproducing the logic of the code, thus the tests does not really test per se.
Hardcode the input and output test values and compare the actual output to the test output.

tests/components/UserActions/QueryList.spec.ts
```diff
@@ -110,6 +112,42 @@ describe('QueryList', () => {
...
+            const button = mock.cmp.element.querySelector<HTMLElement>('.coveo-more-less');
+            if (button) {
+                button.click();
+            }
+
+            const list = mock.env.element.querySelector<HTMLOListElement>('.coveo-list');
+            expect(list.childElementCount).toBe(new Set(TEST_DUPLICATE.map(action => action.query)).size);
+
+            // Check that the order is respected.
+            TEST_DUPLICATE.reverse()
```
louis-bompart: No, you're reproducing the logic of the code, thus the tests does not really test per se.
Hardcode the input and output test values and compare the actual output to the test output.

tests/components/UserActions/ClickedDocumentList.spec.ts
jeremierobert-coveo: The random part prevent depending on a specific string value.

tests/components/UserActions/ClickedDocumentList.spec.ts
jeremierobert-coveo: hardcode ark, but why not 😐

tests/components/UserActions/QueryList.spec.ts
jeremierobert-coveo: ditto

tests/components/UserActions/ClickedDocumentList.spec.ts
louis-bompart: Personally I would rather have documentTestUri that a random, that's some over-engineering imho.
What do you think @tedre191, @erocheleau ?

tests/components/UserActions/ClickedDocumentList.spec.ts
louis-bompart: Well, let's think about what the value of a unit test. It should ensure that the 'black box' that is your code, when given a specific input give you back an expected output. If you repeat the black box in the tests, you're not unit-testing.

tests/components/UserActions/QueryList.spec.ts
louis-bompart: ditto ditto? dittoto ? dittotto ?

tests/components/UserActions/ClickedDocumentList.spec.ts
jeremierobert-coveo: It's not reproduce. It just happen that the reduction to sort thing out is the same.

PR #30, opened 2019-08-23
fix(SFINT-2500): Add ts-es5-istanbul-coverage loader

louis-bompart:
This package will add the istanbul ignore statement where they need to be to circumvent the issue with TypeScript transpilation to ES5.
This will only apply when the code is transpiled for unit tests, not for production.
Also remove all the istanbul ignore comments now that they're not needed anymore.

### DIFF
--- a/config/webpack.config.karma.js
+++ b/config/webpack.config.karma.js
@@ -3,21 +3,42 @@ const path = require('path');
 
 // These modifications are required to have proper coverage with karma-coverage-istanbul-reporter.
 webpackConfig.devtool = 'inline-source-map';
-webpackConfig.module.rules.find(rule => rule.loader === 'ts-loader').options.compilerOptions = {
-    module: 'commonjs',
-    inlineSourceMap: true,
-    sourceMap: undefined,
-    outDir: undefined
-};
-webpackConfig.module.rules.push({
-    enforce: 'post',
-    test: /\.ts$/,
-    loader: 'istanbul-instrumenter-loader',
-    exclude: path.resolve('tests/'),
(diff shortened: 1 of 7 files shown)


config/webpack.config.karma.js
```diff
@@ -3,21 +3,42 @@ const path = require('path');
...
-    outDir: undefined
-};
-webpackConfig.module.rules.push({
-    enforce: 'post',
-    test: /\.ts$/,
-    loader: 'istanbul-instrumenter-loader',
-    exclude: path.resolve('tests/'),
-    query: {
-        esModules: true
+webpackConfig.module.rules = [
```
jeremierobert-coveo: Pourquoi tu push plus la config ?

config/webpack.config.karma.js
```diff
@@ -3,21 +3,42 @@ const path = require('path');
...
+    {
+        test: /\.ts$/,
+        use: [
+            {
+                loader: 'ts-es5-istanbul-coverage'
```
jeremierobert-coveo: Le premier loader deviens l'instrumenteur ?

config/webpack.config.karma.js
louis-bompart: Car ca devenait complexe pour pas vraiment grand chose imo

config/webpack.config.karma.js
louis-bompart: Check: https://www.npmjs.com/package/ts-es5-istanbul-coverage
tl;dr: it runs second. so it instrument the instrumented

PR #38, opened 2019-09-17
Add icons in User Actions Summary SFINT-2368

nathanlb:
Updated CSS so it won't be overwritten in Salesforce.

### DIFF
--- a/src/components/UserActions/QueryList.ts
+++ b/src/components/UserActions/QueryList.ts
@@ -12,12 +12,12 @@ const DEFAULT_TRANSFORMATION = () => (query: string) => {
     icon.classList.add('coveo-row-icon');
     icon.innerHTML = search;
     
-    const span = document.createElement('span');
-    span.classList.add('coveo-content');
-    span.innerHTML = query;
+    const link = document.createElement('a');
+    link.classList.add('coveo-link');
+    link.innerHTML = query;
 
     container.appendChild(icon);
-    container.appendChild(span);
+    container.appendChild(link);
 
     return Promise.resolve(container);
 };
--- a/src/components/UserActions/QueryList.ts
+++ b/src/components/UserActions/QueryList.ts
@@ -102,6 +102,12 @@ export class QueryList extends Component {
         super(element, QueryList.ID, bindings);
 
         this.options = ComponentOptions.initComponentOptions(element, QueryList, options);
+
+        if (!this.options.userId) {
+            this.disable();
+            return;
+        }
+
         this.userProfileModel = get(this.root, UserProfileModel) as UserProfileModel;
         this.userProfileModel.getActions(this.options.userId).then(actions => {
             this.sortedQueryList = [...actions]
--- a/src/components/UserActions/QueryList.ts
+++ b/src/components/UserActions/QueryList.ts
@@ -1,14 +1,25 @@
 import { Component, IComponentBindings, Initialization, ComponentOptions, get, Omnibox, l } from 'coveo-search-ui';
 import { UserProfileModel } from '../../models/UserProfileModel';
 import { ExpandableList } from './ExpandableList';
+import { search } from '../../utils/icons';
 import './Strings';
 
 const DEFAULT_TRANSFORMATION = () => (query: string) => {
+    const container = document.createElement('div');
+
+    const icon = document.createElement('div');
+    icon.classList.add('coveo-inline-item');
+    icon.innerHTML = search;
+    
     const span = document.createElement('span');
     span.classList.add('coveo-content');
+    span.classList.add('coveo-inline-item');
     span.innerHTML = query;
 
-    return Promise.resolve(span);
+    container.appendChild(icon);
+    container.appendChild(span);
+
+    return Promise.resolve(container);
 };
 
 /**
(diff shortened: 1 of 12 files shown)


src/components/UserActions/UserActions.scss
```diff
@@ -84,10 +84,27 @@ $search-ui-dropdown-header-color: #1d4f76;
...
+            text-indent: -1em;
+            padding-left: 1em;
 
             .coveo-list-row > * {
                 display: inline;
             }
+            
+            .coveo-list-row {
+
+                .coveo-row-icon {
```
jeremierobert-coveo:
```
                .coveo-row-icon {
                    margin-left: -2px;
                    & > {
                        svg {
                            height: 1em;
                            width: 1em;
                        }
                    }
                }
```

jeremierobert-coveo: sass let you simplify this way

louis-bompart:
Open discussion, about CSS and the JSUI-Extension and its uses.
When a component looks good in 'standalone', but have some issue in an integration, should we add customization in the integration or update the CSS of the standalone component to 'manage' this.

My point here is, if the CSS is broken by Salesforce, it should be fixed in the Salesforce Integration because search-ui-extensions is integration/platform agnostic IMO, because its goal is to be used by various projects.

erocheleau:
Good question @louis-bompart , my opinion on this is that if Salesforce changes some default styling for example `p { ... }` and that it breaks the component styling, I believe the component could include a "reset css" rule to make it's own `p` standard if it's needed for example.

But for example if the component display breaks because of Salesforce font-size that could be too big and make buttons overflow or something like that, the component is not responsible and should let the products team implements these fixes.

src/components/UserActions/UserActions.scss
```diff
@@ -84,10 +84,26 @@ $search-ui-dropdown-header-color: #1d4f76;
 
         li {
```
erocheleau: @louis-bompart I think that's a good example for your open question.
The component needs specific styling for `li` here. 
So you create a rule with `.class > li` and not just `li` for example which Salesforce may control. That way you risk less conflicts by controlling what you need.

louis-bompart: I agree 👍 
However I'm wondering, if that `text-indent: -1em` etc is because of Salesforce, or not.

jeremierobert-coveo:
Indeed, shielding ourselves with reset css rule is genuinely a good practice but font-size and stuff specific to salesforce env should be done in the salesforce integration.

src/components/UserActions/UserActions.scss
jeremierobert-coveo: What's triggering is more the `text-indent: -1em + padding-left: 1em`.

nathanlb: The indent rules aren't Salesforce specific. I put them there so the text on multi-line items wouldn't start after the icon "bullet" I think it looks nicer this wa but that's open to discussion as well.

nathanlb:
This update isn't really salesforce specific. What is a better way to put it is to say that the issue was noticed due to the context differences found in Salesforce. The icon sizing style rule is there to make sure that this important style is given a high enough priority so that the JSUI-extension package will display as desired in spite of potential context differences.

jeremierobert-coveo:
Screenshot required here 😸

nathanlb:
Without indent:
64884888-c39aff00-d630-11e9-9779-52c168da6e87

With indent:
671325db-d457-4d39-8faf-32654d34bb90

nathanlb:
In this specific instance, the css rule was added because there was no css rule. The html markup for the svg icon contained width and height properties set to 1em. However, those were overridden by the salesforce css.

nathanlb:
I changed the way I styled the whole row to flex. No indent. No padding. 
Made some other stylistic changes based on feedback discussed in Jira.

src/components/UserActions/UserActions.scss
```diff
@@ -83,10 +84,27 @@ $search-ui-dropdown-header-color: #1d4f76;
...
-            margin: 0.25em 0;
+            
+            .coveo-list-row {
+                display: flex;
 
-            .coveo-list-row > * {
-                display: inline;
+                a {
+                    width: 95%;
```
jeremierobert-coveo: why not 100% minus `$icon-size` ?

src/sass/Text.scss
```diff
@@ -9,7 +9,7 @@
 .coveo-content {
     color: black;
     font-size: 1em;
-    font-weight: bold;
+    width: 95%;
```
jeremierobert-coveo: Why not 100% minus $icon-size ?

nathanlb:
Screen Shot 2019-09-19 at 11 38 18 AM

src/components/UserActions/QueryList.ts
```diff
@@ -9,14 +9,15 @@ const DEFAULT_TRANSFORMATION = () => (query: string) => {
     container.classList.add('coveo-list-row');
 
     const icon = document.createElement('div');
+    icon.classList.add('coveo-row-icon');
     icon.innerHTML = search;
     
-    const span = document.createElement('span');
-    span.classList.add('coveo-content');
-    span.innerHTML = query;
+    const link = document.createElement('a');
```
jeremierobert-coveo: Why do we change the span node to a link node ?

nathanlb: Neil requested the change so we could more easily highlight the fact that it was clickable. 
Link to discussion: https://coveord.atlassian.net/browse/SFINT-2368

jeremierobert-coveo: 👍

PR #39, opened 2019-09-19
Sfint 2496 - Added falsey check for userId in DocumentClickedList Component

Neilerino:
Previously the `DocumentClickedList` component would call the `getActions` method for a `UserProfileModel` with a falsey `userId` value when opening a new chat.

This Pull Request adds a falsey check for `userId`. If falsey, the component is disabled.

Added unit test for the falsey check.

Edit: The scope was expanded to fix the same issue in the `QueryList` component and `UserActivity` component.

### DIFF
--- a/src/components/UserActions/ClickedDocumentList.ts
+++ b/src/components/UserActions/ClickedDocumentList.ts
@@ -96,6 +96,11 @@ export class ClickedDocumentList extends Component {
 
         this.options = ComponentOptions.initComponentOptions(element, ClickedDocumentList, options);
 
+        if (!this.options.userId) {
+            this.disable();
+            return;
+        }
+
         this.userProfileModel = get(this.root, UserProfileModel) as UserProfileModel;
 
         this.userProfileModel.getActions(this.options.userId).then(actions => {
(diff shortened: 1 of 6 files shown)


Neilerino:
> No change in the code 👍 It's clean
> However, could you write a more descriptive PR name, only the Jira number is not really great.

Yup!

src/components/UserActions/ClickedDocumentList.ts
```diff
@@ -96,6 +96,11 @@ export class ClickedDocumentList extends Component {
 
         this.options = ComponentOptions.initComponentOptions(element, ClickedDocumentList, options);
 
+        if (!this.options.userId) {
```
jeremierobert-coveo: Check also for `UserActivity` and `QueryList` component. They might suffer the same issue.

tests/components/UserActions/ClickedDocumentList.spec.ts
```diff
@@ -234,6 +234,21 @@ describe('ClickedDocumentList', () => {
         });
     });
 
+    it('Should disable itself when the userId is falsey', () => {
+        let getActionStub: SinonStub<[HTMLElement, ClickedDocumentList], void>;
+        const mock = Mock.advancedComponentSetup<ClickedDocumentList>(
+            ClickedDocumentList,
+            new Mock.AdvancedComponentSetupOptions(null, {userId: null}, env => {
```
jeremierobert-coveo: test also with empty string 😄

PR #40, opened 2019-09-19
fix(SFINT-2563) Add CHROME_BIN export in karma config

louis-bompart:
Simple change in the config, only ensure that the version of chrome being used is the one of puppeteer not the one of the system.
This allow in particular to run tests on WSL.

### DIFF
--- a/config/karma.config.js
+++ b/config/karma.config.js
@@ -1,6 +1,8 @@
 const webpackConfig = require('./webpack.config.karma.js');
 const path = require('path');
 
+process.env.CHROME_BIN = require('puppeteer').executablePath();
+
 var configuration = {
     frameworks: ['jasmine'],
     files: [
--- a/config/karma.config.js
+++ b/config/karma.config.js
@@ -1,6 +1,7 @@
 const webpackConfig = require('./webpack.config.karma.js');
 const path = require('path');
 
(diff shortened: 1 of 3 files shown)


config/karma.config.js
```diff
@@ -1,6 +1,8 @@
 const webpackConfig = require('./webpack.config.karma.js');
 const path = require('path');
 
+process.env.CHROME_BIN = require('puppeteer').executablePath();
```
jeremierobert-coveo: Add ca comment on the why here

config/karma.config.js
```diff
@@ -1,6 +1,9 @@
 const webpackConfig = require('./webpack.config.karma.js');
 const path = require('path');
 
+// Allow crossplatform use of karma-chrome-launcher. See: https://github.com/karma-runner/karma-chrome-launcher#headless-chromium-with-puppeteer
```
Neilerino: In the usage in that readme it has a 
`module.exports = function(config) {
  config.set({
    browsers: ['ChromeHeadless']
  })
}`
Do we need that export here?

louis-bompart: We have it, https://github.com/coveo/search-ui-extensions/blob/9ad796a129b0142c53ae5fbbb61b3a9b5d926294/config/karma.config.js#L50

We just defined the parameter sent to `config.set` as a variable instead of passing it directly.

PR #42, opened 2019-09-25
fix(SFINT-2577): Add karma-webpack-die-hard

louis-bompart:
karma-webpack does not bubble up errors and warning properly, so the build would pass even if errors occurs at the transpilation stage.

abstiles/webpack-karma-die-hard is a simple webpack plugin that ensure that when such error occurs karma is properly notified of it. I added it to the webpack config used for our tests.

You can see the result (i.e. a failed build) [here](https://travis-ci.org/coveo/search-ui-extensions/builds/589582061).

I also fixed the transpilation issues in the tests: two TS6133 errors (variable declared but not used or read) and a TS6192 (All imports in import declaration are unused). This fix the issues raised in the aforementioned failed build.

### DIFF
--- a/tests/components/ResultsFilter/ResultsFilter.spec.ts
+++ b/tests/components/ResultsFilter/ResultsFilter.spec.ts
@@ -2,7 +2,6 @@ import { ResultsFilter, IResultsFilterOptions } from '../../../src/components/Re
 import { Mock, Simulate } from 'coveo-search-ui-tests';
 import { QueryStateModel } from 'coveo-search-ui';
 import { ResultsFilterEvents, IResultsFilterEventArgs } from '../../../src/components/ResultsFilter/Events';
-import { Translation, Language } from '../../../src/utils/translation';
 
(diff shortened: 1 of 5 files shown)


louis-bompart:
f9bc048 fails as expected, now to fix the build.

tests/components/ResultsFilter/ResultsFilter.spec.ts
```diff
@@ -2,7 +2,6 @@ import { ResultsFilter, IResultsFilterOptions } from '../../../src/components/Re
 import { Mock, Simulate } from 'coveo-search-ui-tests';
 import { QueryStateModel } from 'coveo-search-ui';
 import { ResultsFilterEvents, IResultsFilterEventArgs } from '../../../src/components/ResultsFilter/Events';
-import { Translation, Language } from '../../../src/utils/translation';
```
nathanlb: Were these unused?

PR #43, opened 2019-09-25
ci(SFINT-2577): Ensure Prettier run properly on UNIX.

louis-bompart:
The glob was not resolved the same way due to escaping on bash vs Powershell, this is now fixed.
Also now using the standard .prettierrc.js, because .json is not standard.

Bash were the ones with degradation, and thus the CI did not trip on linting ('cause he's on Ubuntu)


louis-bompart:
19acba5 failed as expected.
Now linting the files.

.prettierignore
```diff
@@ -1,10 +1,10 @@
-*.svg
```
Neilerino: this is probably a silly question but what is different here? The changes look the same to me

louis-bompart: No such thing has a silly question ;P 
It's the end of line terminator that have changed: it was a CRLF, it's now a LF, as specified by the .prettierrc.js

Neilerino: Gotcha! Thanks Louis!

package.json
```diff
@@ -23,8 +23,8 @@
         "css": "node ./scripts/css.js",
         "dev": "node ./scripts/dev.server.js",
         "watchTest": "karma start ./config/karma.config.js",
-        "lint": "prettier --check **/*",
-        "lint-fix": "prettier --write **/* ",
+        "lint": "prettier --check \"./**/*\" --config \".prettierrc.js\" --ignore-path \".prettierignore\"",
```
jeremierobert-coveo: Why not targetting the json instead ?

louis-bompart: Because I changed it to a .js to be consistent with the other config files

louis-bompart: Because I changed it to a .js to be consistent with the other config files

jeremierobert-coveo: So where is you change to enable linux ?

louis-bompart: It's the wrapping of the glob (so `**/*` to `\"./**/*\"`)

louis-bompart: It's the wrapping of the glob (so `**/*` to `\"./**/*\"`)

PR #45, opened 2019-10-08
SFINT 2519 - Add the ViewedByCustomer component to every result if the option in UserActions is true

Neilerino:
Added an option to the `UserActions` component to add the ViewedByCustomer component to every result if the `viewedByCustomer` option in `UserAction` is true.

Added typescript unit tests, to test the new private function `showViewedByCustomer` on the `UserAction` component.


src/components/UserActions/UserActions.ts
```diff
@@ -39,6 +50,13 @@ export interface IUserActionsOptions {
      * Default: `User Recent Activity`
      */
     activityLabel: string;
+
+    /**
+     * Weather or not to add the ViewedByCustomer component
```
nathanlb: whether* ;)

tests/components/UserActions/UserActions.spec.ts
```diff
@@ -417,4 +417,80 @@ describe('UserActions', () => {
...
+            const resultArgs = { result: result, item: resultElementRow };
+            Coveo.$$(mock.env.root).trigger(ResultListEvents.newResultDisplayed, resultArgs);
+
+            return delay(() => {
+                expect(mock.cmp.options.viewedByCustomer).toBe(false);
+                expect(resultElementRow.getElementsByClassName('CoveoViewedByCustomer').length).toBe(0);
+            });
+        });
+
+        it('If the div already has ViewedByCustomer component it should not add another', () => {
```
louis-bompart: I'd put this in the describe 'if viewedByCustomer option is set to true', and change the text of it to `it should not add a ViewedByCustomer component if ones is already there`

tests/components/UserActions/UserActions.spec.ts
```diff
@@ -417,4 +417,80 @@ describe('UserActions', () => {
...
+                UserActions,
+                new Mock.AdvancedComponentSetupOptions(null, { userId: 'testUserId' }, env => {
+                    fakeUserProfileModel(env.root, sandbox).getActions.returns(new Promise(() => {}));
+                    return env;
+                })
+            );
+
+            resultElementRow.querySelector('.coveo-result-cell').classList.add('CoveoViewedByCustomer');
+
+            expect(resultElementRow.getElementsByClassName('CoveoViewedByCustomer').length).toBe(1);
```
louis-bompart: I don't think you need this expect.

src/components/UserActions/UserActions.ts
```diff
@@ -231,6 +256,17 @@ export class UserActions extends Component {
         this.element.innerHTML = '';
         this.element.appendChild(element);
     }
+
+    private showViewedByCustomer() {
+        this.bind.onRootElement(ResultListEvents.newResultDisplayed, (args: IDisplayedNewResultEventArgs) => {
+            if (!Boolean(args.item.getElementsByClassName('CoveoViewedByCustomer').length)) {
+                const viewedByCustomerElement = document.createElement('span');
+                new ViewedByCustomer(viewedByCustomerElement, {}, this.bindings, args.result);
```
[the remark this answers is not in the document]
Neilerino: Yeah I like that idea, it'll make this way cleaner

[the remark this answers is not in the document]
louis-bompart: Nah, you're right oh well, let's keep it as it is.

src/components/UserActions/UserActions.ts
```diff
@@ -231,6 +261,31 @@ export class UserActions extends Component {
...
     }
+
+    private showViewedByCustomer() {
+        this.bind.onRootElement(ResultListEvents.newResultDisplayed, (args: IDisplayedNewResultEventArgs) => {
+            if (Boolean(args.item.getElementsByClassName('CoveoViewedByCustomer').length)) {
+                return;
+            }
+            const viewedByCustomerElement = document.createElement('span');
+            new ViewedByCustomer(viewedByCustomerElement, undefined, this.bindings, args.result);
+            args.item.querySelector('.coveo-result-row:last-child .coveo-result-cell').appendChild(viewedByCustomerElement);
```
louis-bompart: Put the selector in a 'constant' (static readonly field of the class)

src/components/UserActions/UserActions.ts
```diff
@@ -231,6 +261,31 @@ export class UserActions extends Component {
...
+        });
+    }
+
+    private tagViewsOfUser() {
+        Coveo.$$(this.root).on('buildingQuery', (e, args) => {
+            try {
+                const createdBy = this.options.record.fields.CreatedBy.value.fields.Email.value;
```
louis-bompart: Hm no-no. This is some Salesforce logic, we don't want that into the JSUI-Extension
Instead of passing the record, only get the 'createdBy' instead

tests/components/UserActions/UserActions.spec.ts
```diff
@@ -417,4 +417,164 @@ describe('UserActions', () => {
...
+                    expect(resultElementRow.getElementsByClassName('CoveoViewedByCustomer').length).toBe(1);
+                });
+            });
+        });
+
+        describe('if the viewedByCustomer option is false', () => {
+            beforeEach(() => {
+                viewedByCustomerOption = false;
+            });
+            it('should not add a ViewedByCustomer Component', () => {
```
louis-bompart: Breakline between 539 and 540, check for good spacing like this across this file :)

tests/components/UserActions/UserActions.spec.ts
```diff
@@ -417,4 +417,164 @@ describe('UserActions', () => {
             });
         });
     });
+
+    describe('tagViewsOfUser', () => {
+        it('should add email to query', () => {
```
louis-bompart: Will re-review this after the Salesforce logic had been extracted.

src/components/UserActions/UserActions.ts
```diff
@@ -269,17 +268,16 @@ export class UserActions extends Component {
             }
             const viewedByCustomerElement = document.createElement('span');
             new ViewedByCustomer(viewedByCustomerElement, undefined, this.bindings, args.result);
-            args.item.querySelector('.coveo-result-row:last-child .coveo-result-cell').appendChild(viewedByCustomerElement);
+            const resultLawRow = args.item.querySelector('.coveo-result-row:last-child .coveo-result-cell');
```
louis-bompart: When I was talking about the selector, I meant just the string `.coveo-result-row:last-child .coveo-result-cell` :P

src/components/UserActions/UserActions.ts
```diff
@@ -63,7 +82,11 @@ export class UserActions extends Component {
         }),
         activityLabel: ComponentOptions.buildStringOption({
             defaultValue: "User's Recent Activity"
-        })
+        }),
+        viewedByCustomer: ComponentOptions.buildBooleanOption({
+            defaultValue: true
+        }),
+        createdBy: ComponentOptions.buildStringOption()
```
jeremierobert-coveo: This option is a duplicate of `userId` 😛

src/components/UserActions/UserActions.ts
```diff
@@ -231,6 +260,30 @@ export class UserActions extends Component {
...
+            const resultLastRow = '.coveo-result-row:last-child .coveo-result-cell';
+            args.item.querySelector(resultLastRow).appendChild(viewedByCustomerElement);
+        });
+    }
+
+    private tagViewsOfUser() {
+        Coveo.$$(this.root).on('buildingQuery', (e, args) => {
+            try {
+                args.queryBuilder.userActions = {
+                    tagViewsOfUser: this.options.createdBy
```
jeremierobert-coveo: Replace `this.option.createBy` by `this.option.userId`. 😸

src/components/UserActions/UserActions.ts
```diff
@@ -231,6 +260,30 @@ export class UserActions extends Component {
...
+                };
+            } catch (e) {
+                this.logger.warn("CreatedBy Email wasn't found", e);
```
jeremierobert-coveo: Maybe make stephanie review this error message 😄

tedre191: No need to review console error messages, but I agree that it shouldn't say "createdby.email" :P

PR #46, opened 2019-10-10
Architecture Design Record for the showViewedByCustomer method in UserActions

Neilerino:
This is an ADR for the showViewedByCustomer  method in: https://coveord.atlassian.net/browse/SFINT-2519

As this is the first ADR in this repo, please give feedback on:
File Path, File Name, File content, File Template, anything else you can think of

My idea behind the file path was that as we create ADRs for each new user story they will start to pile up. So having one folder for each component (following the same folder structure as the src/) would make it easy to find the one your looking for when you want to find one in the future.

I wasn't super sure about the file name. IMO it should be something that clearly explains what the user story impacts, so that you don't need to root around in each file when you go looking for something. This is why I'm not a big fan of the `ADR-####` file name convention I've seen brought up a few times online. Not sure about having the JIRA in the file name either, I could go either way on that one.

For the template I based it on the one in the link below, but I made it look a bit prettier, and moved some sections into others.  https://github.com/joelparkerhenderson/architecture_decision_record/blob/master/adr_template_madr.md

Also let me know if I should add details anywhere. I'm not totally sure how to do this 😅


nathanlb:
Looks good! Could we also add links to these ADRs in the JIRAs or PRs directly? I feel like they're already a source of info on past changes and contain quite a bit of insight with discussions and commit history. Adding this kind of file would just complete the info and make everything accessible from one spot.

louis-bompart:
In the future, we will commit ADR with the PR themselves.
Its just that we're trying stuff for now :P

PR #48, opened 2019-10-15
fix(SFINT-2612): Remove duplicate options in UserActions

louis-bompart:
@jeremierobert-coveo pointed out that we don't need another options for the viewedBy and can use the userId. This does that.

### DIFF
--- a/tests/components/UserActions/UserActions.spec.ts
+++ b/tests/components/UserActions/UserActions.spec.ts
@@ -422,7 +422,7 @@ describe('UserActions', () => {
         it('should add email to query', () => {
             const mock = Mock.advancedComponentSetup<UserActions>(
                 UserActions,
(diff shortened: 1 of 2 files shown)


tests/components/UserActions/UserActions.spec.ts
```diff
@@ -434,7 +434,7 @@ describe('UserActions', () => {
             Coveo.$$(mock.env.root).trigger('buildingQuery', queryArgs);
 
             return delay(() => {
-                expect(queryData.queryBuilder.userActions.tagViewsOfUser).toBe('test@email.com');
```
jeremierobert-coveo: Why changing this value ?

louis-bompart: Merged the two field, addressed.

louis-bompart:
> This is really clean, sorry I missed this when I was working on it.

No trouble I missed it too xD

PR #50, opened 2019-10-22
Show a contact us message when the User Actions is not enabled SFINT-2570

nathanlb:
# What has changed?
- When a 404 status is returned from a user clicking UserActions without having it enabled in the org, a new [Feature is Not Activated](https://docs.coveo.com/en/1085/coveo-for-salesforce/the-feature-is-not-activated-for-your-organization) message is shown.

# How was it tested?
[X] Unit tests
[X] Manu SAN-Check

# Preview
Screen Shot 2019-10-22 at 2 21 30 PM

### DIFF
--- a/src/components/UserActions/Strings.ts
+++ b/src/components/UserActions/Strings.ts
@@ -2,6 +2,7 @@ import { Translation, Language } from '../../utils/translation';
 
 Translation.register(Language.English, {
     UserActions_no_actions: 'No actions available for this user',
+    UserActions_enable_prompt: 'The User Action feature isn’t enabled for you account.\nPlease contact Coveo Support.',
 
     QueryList_more: 'Show More',
     QueryList_less: 'Show Less',
--- a/src/components/UserActions/Strings.ts
+++ b/src/components/UserActions/Strings.ts
@@ -2,7 +2,7 @@ import { Translation, Language } from '../../utils/translation';
 
 Translation.register(Language.English, {
     UserActions_no_actions: 'No actions available for this user',
-    UserActions_enable_prompt: 'The User Action feature isn’t enabled for you account.\nPlease contact Coveo Support.',
+    UserActions_enable_prompt: 'The User Action feature is not activated for your organization.\nTo activate it, contact Coveo Support.',
 
     QueryList_more: 'Show More',
     QueryList_less: 'Show Less',
--- a/src/components/UserActions/UserActions.ts
+++ b/src/components/UserActions/UserActions.ts
@@ -143,9 +143,7 @@ export class UserActions extends Component {
             (get(this.root, UserProfileModel) as UserProfileModel)
                 .getActions(this.options.userId)
                 .then(actions => (actions.length > 0 ? this.render() : this.renderNoActions()))
-                .catch(
-                    (e) => { e.statusCode === 404 ? this.renderEnablePrompt() : this.renderNoActions() }
-                );
+                .catch(e => { e.statusCode === 404 ? this.renderEnablePrompt() : this.renderNoActions() });
 
             this.bindings.usageAnalytics.logCustomEvent({ name: 'openUserActions', type: 'User Actions' }, {}, this.element);
             this.root.classList.add(UserActions.USER_ACTION_OPENED);
--- a/src/components/UserActions/UserActions.ts
+++ b/src/components/UserActions/UserActions.ts
@@ -143,7 +143,9 @@ export class UserActions extends Component {
             (get(this.root, UserProfileModel) as UserProfileModel)
                 .getActions(this.options.userId)
                 .then(actions => (actions.length > 0 ? this.render() : this.renderNoActions()))
-                .catch(e => { e.statusCode === 404 ? this.renderEnablePrompt() : this.renderNoActions() });
+                .catch(e => {
+                    e.statusCode === 404 ? this.renderEnablePrompt() : this.renderNoActions();
+                });
 
             this.bindings.usageAnalytics.logCustomEvent({ name: 'openUserActions', type: 'User Actions' }, {}, this.element);
             this.root.classList.add(UserActions.USER_ACTION_OPENED);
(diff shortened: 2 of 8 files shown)


jeremierobert-coveo:
Does the `contact Coveo Support` clikable ?

jeremierobert-coveo:
The message is only shown when you click on the user action button ?

src/components/UserActions/Strings.ts
```diff
@@ -2,6 +2,7 @@ import { Translation, Language } from '../../utils/translation';
 
 Translation.register(Language.English, {
     UserActions_no_actions: 'No actions available for this user',
+    UserActions_enable_prompt: 'The User Action feature is not activated for your organization.\nTo activate it, contact Coveo Support.',
```
jeremierobert-coveo: Does DOC have pass over this string ?

nathanlb:
> The message is only shown when you click on the user action button ?

Yes. otherwise it shows normal search results.

src/components/UserActions/UserActions.ts
```diff
@@ -143,7 +143,9 @@ export class UserActions extends Component {
             (get(this.root, UserProfileModel) as UserProfileModel)
                 .getActions(this.options.userId)
                 .then(actions => (actions.length > 0 ? this.render() : this.renderNoActions()))
-                .catch(() => this.renderNoActions());
+                .catch(e => {
+                    e.statusCode === 404 ? this.renderEnablePrompt() : this.renderNoActions();
+                });
```
jeremierobert-coveo: Do you think you can simplify this ?

like this maybe ?
```javascript
.catch(e => (e.statusCode === 404 ? this.renderEnablePrompt() : this.renderNoActions()));
```

nathanlb:
> Does the `contact Coveo Support` clikable ?

No Steph didn't mention a link but it could...

tests/components/UserActions/UserActions.spec.ts
```diff
@@ -190,6 +191,24 @@ describe('UserActions', () => {
         expect(hideSpy.called).toBe(true);
     });
 
+    it('should show a message when user actions is not enabled', () => {
```
jeremierobert-coveo: Nice test 😄 !

Clean and short.

tests/components/UserActions/UserActions.spec.ts
```diff
@@ -204,7 +223,7 @@ describe('UserActions', () => {
 
         return delay(() => {
             expect(mock.cmp.element.querySelector<HTMLElement>('.coveo-no-actions')).not.toBeNull();
-            expect(mock.cmp.element.querySelector<HTMLElement>('.coveo-no-actions').innerText).toBe('No actions available for this user');
+            expect(mock.cmp.element.querySelector<HTMLElement>('.coveo-no-actions').innerText).toBe(l(`${UserActions.ID}_no_actions`));
```
jeremierobert-coveo: Nice catch

src/components/UserActions/UserActions.ts
nathanlb: That's what i had :(

nathanlb: Linter says no

src/components/UserActions/Strings.ts
nathanlb: Yes Steph checked it out

src/components/UserActions/UserActions.ts
louis-bompart: 👍 if this is really a pain (I, personally, don't think so), changes the rules of prettier, in another PR.

src/components/UserActions/UserActions.ts
```diff
@@ -143,7 +143,9 @@ export class UserActions extends Component {
             (get(this.root, UserProfileModel) as UserProfileModel)
                 .getActions(this.options.userId)
                 .then(actions => (actions.length > 0 ? this.render() : this.renderNoActions()))
-                .catch(() => this.renderNoActions());
+                .catch(e => {
+                    e.statusCode === 404 ? this.renderEnablePrompt() : this.renderNoActions();
```
louis-bompart: mhhh.... it's between curly braces and the then l145 is between 'simple' braces '(', is that normal?

src/components/UserActions/UserActions.ts
```diff
@@ -264,6 +266,16 @@ export class UserActions extends Component {
         this.element.appendChild(element);
     }
 
+    private renderEnablePrompt() {
+        const element = document.createElement('div');
+        element.classList.add('coveo-no-actions');
+        element.innerText = l(`${UserActions.ID}_enable_prompt`);
+        element.style.padding = '1.5em';
```
louis-bompart: Why not apply the padding in CSS on the 'coveo-no-actions' class?

nathanlb: This padding is a reduction from the 3em padding already in the css class to allow the message to be displayed cleanly on two lines instead of three. I assumed the original padding value had some thought behind it so I didn't want to overwrite it everywhere.

src/components/UserActions/UserActions.ts
nathanlb: Nice catch! The linter allows this to be on one line 👍

src/components/UserActions/UserActions.ts
louis-bompart: 🤔 hm maybe a custom css class and some css then? It is my opinion here, but I'm not a huge fan of styling an element over the style attribute and prefer to avoid it when reasonable/possible.

nathanlb: I agree that it isn't ideal. I just wanted to avoid essentially duplicating the class. On the other hand, the class name "no-actions" isn't super appropriate either... Ok I'll change it.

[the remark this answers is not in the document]
nathanlb: Good tip but I opted for replacing the no-actions class with a new one. I hadn't looked before but the no-actions class only has two properties, including the padding.

PR #52, opened 2019-10-30
Create new star rating widget SFINT-2649

nathanlb:
# Overview
Created a new StarRating typescript component that takes two fields **rating**, **numberOfRatings** and displays a star rating with the given number of ratings

# Tests
[X] Unit tests

# Preview
```html
<div class="CoveoStarRating" data-rating="@rating" data-number-of-ratings="@numratings" data-rating-scale="10">
```
Screen Shot 2019-10-30 at 10 15 26 AM


src/components/StarRating/StarRating.ts
```diff
@@ -0,0 +1,63 @@
+import { $$, Utils, Component, Initialization, ComponentOptions, IComponentBindings } from 'coveo-search-ui';
+import { star as starIcon } from '../../utils/icons';
+
+export interface IStarRatingOptions {
+    /** Specifies the rating to be displayed as stars. */
+    rating: number;
```
tedre191: that should probably be a "field" instead of a number ? How will you get that number otherwise from the index ?

src/components/StarRating/StarRating.ts
```diff
@@ -0,0 +1,63 @@
...
+    /** Specifies the number to be displayed indicating the number of reviews. */
+    numberOfRatings?: number;
```
tedre191: same here

src/components/StarRating/StarRating.ts
```diff
@@ -0,0 +1,63 @@
...
+    };
+
+    /**
+     * Creates a new `StarRating` component.
+     * @param element The HTMLElement on which to instantiate the component.
+     * @param options The options for the `StarRating` component.
+     * @param bindings The bindings that the component requires to function normally. If not set, these will be
+     * automatically resolved (with a slower execution time).
+     */
+    constructor(public element: HTMLElement, public options: IStarRatingOptions, public bindings?: IComponentBindings) {
```
tedre191: you should probably get the "result" bindings and all here (check the ViewedByCustomer constructor for reference). This is also linked to my previous comment about passing "fields" instead of direct value :)

src/components/StarRating/StarRating.ts
```diff
@@ -0,0 +1,63 @@
+import { $$, Utils, Component, Initialization, ComponentOptions, IComponentBindings } from 'coveo-search-ui';
+import { star as starIcon } from '../../utils/icons';
+
+export interface IStarRatingOptions {
```
tedre191: Should we support different "range" of score ? Let's say somebody has a 0-10 scoring system that we could map back to a five star system ?

src/components/StarRating/StarRating.ts
```diff
@@ -0,0 +1,63 @@
...
+    private renderStar(element: HTMLElement, isChecked: boolean, value: number) {
+        const star = $$('span', { className: 'coveo-star-rating-star' }, starIcon);
+        star.children()[0].classList.add('coveo-star-rating-star-svg');
+        element.appendChild(star.el);
+        star.toggleClass('coveo-star-rating-star-active', isChecked);
+    }
+
+    private renderNumberOfReviews(element: HTMLElement, value: number) {
+        const numberString = $$('span', { className: 'coveo-star-rating-label' });
+        value > 0 ? (numberString.el.innerText = `(${value})`) : (numberString.el.innerText = 'No Ratings');
```
tedre191: The "No Ratings" string should be localized using the search-ui's localization system :)

src/components/StarRating/StarRating.scss
```diff
@@ -0,0 +1,31 @@
+$color-placeholder-star: #e1e1e1;
+$color-active-star: #d9af00;
+$color-number-reviews: gray;
+
+.CoveoStarRating {
+    display: flex;
+    flex-direction: row;
+    flex-wrap: nowrap;
```
Neilerino: I don't thinnnk you need these two lines. Both of those are the default attributes for `display:flex` iirc 🤔

src/components/StarRating/StarRating.scss
```diff
@@ -0,0 +1,31 @@
...
+    .coveo-star-rating-star-svg {
+        color: $color-active-star;
+    }
+}
+
+.coveo-star-rating-label {
+    margin-left: 5px;
+    font-size: 11px;
+    color: $color-number-reviews;
+}
```
Neilerino: You can simplify this by using extending a base `.coveo-star-rating` class like this:
`.coveo-star-rating-star {
   &-svg {
      stuff here
   }
   other stuff here
}`

src/components/StarRating/StarRating.scss
nathanlb: You are correct sir

img/star.svg
```diff
@@ -0,0 +1,5 @@
+<svg alt="Star" focusable="false" enable-background="new 0 0 16 16" viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg">
```
jeremierobert-coveo: set a default height and width so in case the css is not loaded we have something not so bad

src/components/StarRating/StarRating.scss
```diff
@@ -0,0 +1,31 @@
...
+
+.coveo-star-rating-star-active {
+    .coveo-star-rating-star-svg {
+        color: $color-active-star;
+    }
+}
+
+.coveo-star-rating-label {
+    margin-left: 5px;
+    font-size: 11px;
```
jeremierobert-coveo: Where thosemagic number comes from ?

src/components/StarRating/StarRating.scss
```diff
@@ -0,0 +1,31 @@
...
+$color-active-star: #d9af00;
+$color-number-reviews: gray;
+
+.CoveoStarRating {
+    display: flex;
+    flex-direction: row;
+    flex-wrap: nowrap;
+}
+
+.coveo-star-rating-star + .coveo-star-rating-star {
```
Neilerino: Why are you using the adjacent selector on a class with itself?

src/components/StarRating/StarRating.ts
nathanlb: What do you mean by "field"? The value is passed like in the example and number gets converted fine.

src/components/StarRating/StarRating.ts
nathanlb: I did it this way to give more flexibility to the user. Similarly to how the other ResultRating was unusable in our context, tying it to a result field might make it inadequate for another use case no?

tests/components/StarRating/StarRating.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+            it('should display a label with "No Ratings" shown when there are none', () => {
+                const testOptions: IStarRatingOptions = { rating: 0 };
+                const testSubject = Mock.basicComponentSetup<StarRating>(StarRating, testOptions);
+                const testLabel = testSubject.cmp.element.lastElementChild;
+                expect(testLabel.className).toEqual(STAR_LABEL_CSS_CLASS);
+                expect(testLabel.textContent).toEqual('No Ratings');
+            });
+
+            it('should display the number when they are provided', () => {
+                const testNumRatings = 69;
```
jeremierobert-coveo: This way I'm sure you cover everything 😉

src/components/StarRating/StarRating.scss
nathanlb: More pixels looks less good. less pixels looks less good. magic! :P
What kind of correlation are we usually looking for?

src/components/StarRating/StarRating.ts
erocheleau: What do you envision as another use case? Other than that being bound to a field?

src/components/StarRating/StarRating.scss
erocheleau: May be something you want to enforce though in the case where other CSS might be laying around in the page?
I would prefer being explicit here.

src/components/StarRating/StarRating.ts
nathanlb: Maybe I'm not understanding what you're saying. Being bound to a field means it would be bound to, let's say, @sfccrzproductratingc ?

src/components/StarRating/StarRating.scss
erocheleau: why is the font-size 11px?
why is it 5px of margin-left?

You could use `em` instead perhaps?

tests/components/StarRating/StarRating.spec.ts
erocheleau: what about 70? Do you cover 70? :D

src/components/StarRating/StarRating.scss
erocheleau: That's actually smart but a costly selector in CSS :) don't know what we want to do with it.

src/components/StarRating/StarRating.scss
nathanlb: sure we can. It would make sense to size the stars according to em too then. 11px was chosen because it looked best with a fixed star size fo 16px.

src/components/StarRating/StarRating.ts
erocheleau: Yes, you don't want to have to write

<div class="CoveoStarRating" data-rating="4" data-number-of-rating="10">

in your result templates right? ;)

tests/components/StarRating/StarRating.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+const CONTAINER_CSS_CLASS = 'CoveoStarRating';
+const STAR_CSS_CLASS = 'coveo-star-rating-star';
+const ACTIVE_STAR_CSS_CLASS = 'coveo-star-rating-star-active';
+const STAR_LABEL_CSS_CLASS = 'coveo-star-rating-label';
+
+const getActiveStars = (starRatingElement: HTMLElement) => {
+    const children = starRatingElement.children;
+    let numStars = 0;
+    let numActiveStars = 0;
+    for (let i = 0; i < children.length; i++) {
```
Neilerino: Can you use `forEach` here instead of looping and using the index?

tests/components/StarRating/StarRating.spec.ts
nathanlb: Screen Shot 2019-10-30 at 4 41 50 PM

src/components/StarRating/StarRating.ts
```diff
@@ -0,0 +1,63 @@
...
+    constructor(public element: HTMLElement, public options: IStarRatingOptions, public bindings?: IComponentBindings) {
+        super(element, StarRating.ID);
+        this.options = ComponentOptions.initComponentOptions(element, StarRating, options);
+        if (!Utils.isNullOrUndefined(options.rating)) {
+            this.renderComponent(element, options);
+        }
+    }
+
+    private renderComponent(element: HTMLElement, options: IStarRatingOptions) {
+        for (let starNumber = 1; starNumber <= 5; starNumber++) {
```
erocheleau: I would set the maxStar as an option here.

src/components/StarRating/StarRating.scss
nathanlb: no worries I know what you're talking about. Like i said elsewhere, i just copied the css and changes the class names not wanting to change what isn't broke

tests/components/StarRating/StarRating.spec.ts
nathanlb: Haha tried that at first but `children` isn't an array so it doesn't have a forEach. It's an HTMLCollection

src/components/StarRating/StarRating.ts
nathanlb: I see I thought as it is putting `data-rating="@sfccrzproductratingc"` would work. I was under the impression the binding alternative would tie it to that explicitly named field. So the other use cases I was thinking of was in case anyone wanted to use this component in a non-ccrz context

louis-bompart: Do we want to allow the customization of the star icon?

tests/components/StarRating/StarRating.spec.ts
louis-bompart: What about... `Number.MAX_SAFE_INTEGER`?

tests/components/StarRating/StarRating.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+        describe('The label showing number of ratings', () => {
+            it('should display a label with "No Ratings" shown when there are none', () => {
+                const testOptions: IStarRatingOptions = { rating: 0 };
+                const testSubject = Mock.basicComponentSetup<StarRating>(StarRating, testOptions);
+                const testLabel = testSubject.cmp.element.lastElementChild;
+                expect(testLabel.className).toEqual(STAR_LABEL_CSS_CLASS);
+                expect(testLabel.textContent).toEqual('No Ratings');
+            });
+
+            it('should display the number when they are provided', () => {
```
louis-bompart: Add some breaklines in there:)
Like between
`const testSubject = Mock.basicComponentSetup<StarRating>(StarRating, testOptions);`
`const testLabel = testSubject.cmp.element.lastElementChild;`

It would mark the separation between the [exercise and the verify test steps](https://thoughtbot.com/blog/four-phase-test)

src/components/StarRating/StarRating.ts
louis-bompart: I found really odd how you built this ternary, I would have done it like this, mostly because I most often saw it used like this.
```suggestion
        numberString.el.innerText = value > 0 ? value.toString() : 'No Ratings';
```

Although you would have to modify the No Ratings string to include the localization that André mentioned

src/components/StarRating/StarRating.ts
```diff
@@ -0,0 +1,63 @@
...
+            this.renderStar(element, starNumber <= options.rating, starNumber);
+        }
+        this.renderNumberOfReviews(element, options.numberOfRatings);
+    }
+
+    private renderStar(element: HTMLElement, isChecked: boolean, value: number) {
+        const star = $$('span', { className: 'coveo-star-rating-star' }, starIcon);
+        star.children()[0].classList.add('coveo-star-rating-star-svg');
+        element.appendChild(star.el);
```
louis-bompart: Why `element.appendChild(star.el);` is before `star.toggleClass('coveo-star-rating-star-active', isChecked);`

Would it not make sense to compute all the CSS then append it ?

nathanlb: Yeah you know I would think that too. But i've seen it done in random order all over the codebase. I assume the result is the same. Just logically odd.

src/components/StarRating/StarRating.ts
nathanlb: Yeah yours is better. I just used if statement logic

tests/components/StarRating/StarRating.spec.ts
nathanlb: :)

tests/components/StarRating/StarRating.spec.ts
erocheleau: You can convert it to array with a `[...htmlCollection]` if necessary but that's not necessary I think :)

nathanlb: Good to know!

src/components/StarRating/StarRating.ts
jeremierobert-coveo: Actually the result is not the same perf wise. BUT it can be done this way if you append the element on an element not on the dom yet.

nathanlb: Right yes of course

src/components/StarRating/StarRating.ts
nathanlb: Done

nathanlb: It rounds down though. Could try to have the stars fill in partially with some fancy css but I don't think that's needed for now.

PR #54, opened 2019-12-10
fix(SFINT-2686): Change error to warn

louis-bompart:
Changed the error to a warn to avoid the crash of the page and just to be a bit less brutal.

### DIFF
--- a/src/components/ViewedByCustomer/ViewedByCustomer.ts
+++ b/src/components/ViewedByCustomer/ViewedByCustomer.ts
@@ -55,7 +55,9 @@ export class ViewedByCustomer extends Component {
     ) {
         super(element, ViewedByCustomer.ID, bindings);
         if (this.root.getElementsByClassName(Component.computeCssClassNameForType(UserActions.ID)).length === 0) {
-            throw new Error('No UserActions component found on the page template.');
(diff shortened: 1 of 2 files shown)


src/components/ViewedByCustomer/ViewedByCustomer.ts
```diff
@@ -55,7 +55,9 @@ export class ViewedByCustomer extends Component {
     ) {
         super(element, ViewedByCustomer.ID, bindings);
         if (this.root.getElementsByClassName(Component.computeCssClassNameForType(UserActions.ID)).length === 0) {
-            throw new Error('No UserActions component found on the page template.');
+            this.logger.warn(
+                'ViewedByCustomer component has been detected without a UserAction components, you may encounter issues with the former.'
```
nathanlb: Doc approved?

Stremblay2: The ViewedByCustomer component has been detected without a UserActions component. You may encounter issues with the former.
(Then doc approved :D)

louis-bompart: Done!

PR #56, opened 2019-12-16
Pass init options for UserActions sub components. SFINT-2759

lbergeron:
Custom init options were ignored for the UserActions sub components.

### DIFF
--- a/tests/components/UserActions/UserActions.spec.ts
+++ b/tests/components/UserActions/UserActions.spec.ts
@@ -190,20 +190,21 @@ describe('UserActions', () => {
 
     it('should pass custom init options to each of the sub components', () => {
         const FAKE_USER_ID = 'someUserId' + Math.random();
-        const initOptions = {} as any;
-        initOptions[QueryList.ID] = {
-            listLabel: 'Custom Query List Title',
-            numberOfItems: 1
-        };
-        initOptions[ClickedDocumentList.ID] = {
-            listLabel: 'Custom Clicked Document List Title',
-            template: 'Custom Template',
-            numberOfItems: 2
-        };
-        initOptions[UserActivity.ID] = {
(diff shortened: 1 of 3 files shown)


tests/components/UserActions/UserActions.spec.ts
```diff
@@ -188,6 +188,49 @@ describe('UserActions', () => {
         });
     });
 
+    it('should pass custom init options to each of the sub components', () => {
+        const FAKE_USER_ID = 'someUserId' + Math.random();
+        const initOptions = {} as any;
```
louis-bompart: Why `as any` ?

jeremierobert-coveo: here I think you can declare the whole object in a single run. This will provide intellisense and will prevent the use of any 😄 .

lbergeron: Without the `as any`, TypeScript was complaining when trying to execute `initOptions[component]` where `component` is of type String.

lbergeron: When declaring the whole object in one shot, I had this compiler error.

    No index signature with a parameter of type 'string' was found on type

jeremierobert-coveo: Yes you need to cast your object in a dict like that `as { [key: string]: any }` to prevent this error this way you said explicitly that you play with a Dict 😃 .

lbergeron: 🤦‍♂ Thanks

jeremierobert-coveo: Obscure typings for the win 🐈 .

lbergeron: Specified `initOptions` as a dictionary instead of `any`.

PR #57, opened 2020-01-06
Added type button to the More/Less button in expandable list

mikegron:
Without the `type=button`  attribute, the default behaviour of the button is to submit the parent form, if there is any. That was a problem in ServiceNow since the whole page is a huge form, clicking on Show More/Less would submit the page and redirect the user. 

SNOW-240


jeremierobert-coveo:
Oh no that's bad 😱

PR #58, opened 2020-01-27
fix(SFINT-2851): BREAKING Insert ViewedByCustomer in a new row

louis-bompart:
Jira: https://coveord.atlassian.net/browse/SFINT-2851

### The issue

If the last cell/row of the result is already quite full, the Viewed By Customer would break in two lines and that didn't render very well.

### Previous behavior

When 'auto-appending' the ViewedByCustomer in the result, append the ViewedByCustomer in the first cell of the last row of the result frame.

### New behavior

When 'auto-appending' the ViewedByCustomer in the result, find the last row of the result row and create a new 'sibling' row in which there's a cell that include the ViewedByCustomer.

### Screenshot
Top, before. Bottom, after.
image

### Tests

- [X] UT
- [X] SAN Check

### DIFF
--- a/src/components/UserActions/UserActions.ts
+++ b/src/components/UserActions/UserActions.ts
@@ -288,20 +288,10 @@ export class UserActions extends Component {
             if (Boolean(args.item.getElementsByClassName('CoveoViewedByCustomer').length)) {
                 return;
             }
-            const viewedByCustomerRow = document.createElement('div');
-            viewedByCustomerRow.classList.add('coveo-result-row');
-
-            const viewedByCustomerCell = document.createElement('div');
-            viewedByCustomerCell.classList.add('coveo-result-cell');
-
-            const viewedByCustomerElement = document.createElement('span');
-            new ViewedByCustomer(viewedByCustomerElement, undefined, this.bindings, args.result);
-
-            viewedByCustomerCell.appendChild(viewedByCustomerElement);
-            viewedByCustomerRow.appendChild(viewedByCustomerCell);
-
             const resultLastRow = '.coveo-result-row:last-child';
-            args.item.querySelector(resultLastRow).parentNode.appendChild(viewedByCustomerElement);
+            args.item
+                .querySelector(resultLastRow)
+                .parentNode.appendChild(ViewedByCustomer.getViewedByCustomerResultRowDom(this.bindings, args.result));
         });
     }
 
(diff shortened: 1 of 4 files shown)


src/components/UserActions/UserActions.ts
```diff
@@ -288,10 +288,20 @@ export class UserActions extends Component {
             if (Boolean(args.item.getElementsByClassName('CoveoViewedByCustomer').length)) {
                 return;
             }
+            const viewedByCustomerRow = document.createElement('div');
```
jeremierobert-coveo: That seems a bit wrong that the user action is handling the viewed by custom logic no ?

nathanlb: This is how it was before though. We kindof need a "parent" element to insert the component in the result div right?
I'm curious though, what happens if users place it elsewhere in the result templates? Does this overwrite it?

louis-bompart: I think that's another question that one could have (i.e. if we were to change that, that's yet another story).

Also, I invite you to check the Jira linked to the addition of this feature: https://coveord.atlassian.net/browse/SFINT-2519, it seems to me that it was required.

jeremierobert-coveo: You can make a function on the viewedByCustomer side where you generate the required dom, no ?

louis-bompart: No, it's a result template component it cannot add itself to the result template (and all the others)

Also, that's quite a bigger change that you're suggesting, let's discuss about it off GitHub.

louis-bompart: > This is how it was before though. We kindof need a "parent" element to insert the component in the result div right?
> I'm curious though, what happens if users place it elsewhere in the result templates? Does this overwrite it?
@nlafrance-coveo
No, if you turn this option at on, it will add the ViewedByCustomer a the bottom of every result templates that doesn't have it already.

louis-bompart: Cleared misunderstanding with @jeremierobert-coveo which resulted in https://github.com/coveo/search-ui-extensions/pull/58/commits/7d6d02d09f344ffaad810ba8e5c4e71f1e51e5ca

@jeremierobert-coveo , if all green please approve 👍 
Thanks again for taking the time to get to the bottom of this quiproquo

louis-bompart: Cleared misunderstanding with @jeremierobert-coveo which resulted in https://github.com/coveo/search-ui-extensions/pull/58/commits/7d6d02d09f344ffaad810ba8e5c4e71f1e51e5ca

@jeremierobert-coveo , if all green please approve 👍 
Thanks again for taking the time to get to the bottom of this quiproquo

PR #59, opened 2020-02-19
[SFINT-2847] Viewed by customer is no longer automagicly injected in a table layout

jeremierobert-coveo:
This really looks odd. That's why we removed it.

### DIFF
--- a/pages/index.html
+++ b/pages/index.html
@@ -3,11 +3,11 @@
     <head>
         <meta charset="UTF-8" />
         <meta name="viewport" content="width=device-width, height=device-height" />
-        <title>Title</title>
-        <link rel="stylesheet" href="./css/CoveoFullSearch.css" />
-        <link rel="stylesheet" href="./css/CoveoJsSearchExtensions.css" />
-        <script src="http://localhost:8080/js/CoveoJsSearch.Lazy.js"></script>
-        <script src="commonjs/CoveoJsSearchExtensions.js"></script>
+        <title>Development</title>
+        <link rel="stylesheet" href="../css/CoveoFullSearch.css" />
+        <link rel="stylesheet" href="../css/CoveoJsSearchExtensions.css" />
+        <script src="../js/CoveoJsSearch.Lazy.js"></script>
+        <script src="../commonjs/CoveoJsSearchExtensions.js"></script>
         <script>
             let attachedIds = [];
             document.addEventListener('DOMContentLoaded', function() {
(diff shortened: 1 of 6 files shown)


pages/index.html
```diff
@@ -3,11 +3,11 @@
     <head>
         <meta charset="UTF-8" />
         <meta name="viewport" content="width=device-width, height=device-height" />
-        <title>Title</title>
-        <link rel="stylesheet" href="./css/CoveoFullSearch.css" />
-        <link rel="stylesheet" href="./css/CoveoJsSearchExtensions.css" />
-        <script src="http://localhost:8080/js/CoveoJsSearch.Lazy.js"></script>
```
nathanlb: hmm

jeremierobert-coveo: Yeah that was wrong

pages/index.html
```diff
@@ -38,6 +38,9 @@
                         getValues: function() {
                             return attachedIds;
                         }
+                    },
+                    UserActions: {
+                        userId: 'test'
```
nathanlb: This too perhaps?

jeremierobert-coveo: Require to test stuff 😄

nathanlb: Oh i forget is this file just for running the dev server?

jeremierobert-coveo: Yeah 😄

nathanlb: oups!

src/components/UserActions/UserActions.ts
```diff
@@ -306,6 +315,16 @@ export class UserActions extends Component {
             }
         });
     }
+
+    private inferResultListLayout(): ResultLayoutType {
+        const resultLists = this.root.querySelectorAll<HTMLElement>(`${Component.computeSelectorForType(ResultList.ID)}:not(.coveo-hidden)`);
+        const resultListLayoutTypes = [ResultLayoutType.CARD, ResultLayoutType.TABLE, ResultLayoutType.LIST] as string[];
+
+        if (resultLists.length > 0 && resultListLayoutTypes.indexOf(resultLists[0].dataset.layout) !== -1) {
```
lbergeron: You are checking the first result list that is visible. Maybe it is far-fetched, but suppose you have two result lists visible at the same time. The first one has the LIST template and the second one has the TABLE template. Since you are checking the layout for the first visible list, it will still append the "Viewed by Customer" element in the result list with the TABLE template.

Instead, can you check the layout on the first `CoveoResultList` parent?

jeremierobert-coveo: Sadly no, that was my first plan but sadly it's not possible. the dom note given in the handler is not yet attached to the dom.

PR #60, opened 2020-02-24
SFINT-2878 Added Cypress functional tests

jeremierobert-coveo:
I added Cypress to run functional test in the repo.

Here some screenshot.

image

And here too

image


louis-bompart:
Can you split the PR in two?
- The setup
- The tests

So we can distinguish the setup and the tests in our review (and upcoming /blame :P )

cypress.json
```diff
@@ -0,0 +1,11 @@
+{
+    "baseUrl": "http://localhost:8080",
```
louis-bompart: https?

cypress.json
```diff
@@ -0,0 +1,11 @@
...
+    "reporter": "spec",
+    "viewportHeight": 600,
```
louis-bompart: Maybe FHD?

cypress.json
jeremierobert-coveo: nope, c'est discourage by cypress

cypress.json
jeremierobert-coveo: What would be the gain ?

jeremierobert-coveo:
imo, the setup is pointless with at least a single test to test the test setup.
And those User actions tests where quick and simple to write.

cypress.json
louis-bompart: I see magic numbers I propose new ones :P 
1920*1080 seems closer to real life application to me.

louis-bompart:
> imo, the setup is pointless with at least a single test to test the test setup.
> And those User actions tests where quick and simple to write.

The thing is if your tests become an issue, we will have a single commit to revert without having to untangle what to keep, what to ditch.

It keeps things contained. Like, your setup should not be hard-linked with a feature

jeremierobert-coveo:
Well I'll need to add a dummy then 
image

louis-bompart:
> Well I'll need to add a dummy then
> image

Yep, do a dummy, remove it in the PR that include the first FT

jeremierobert-coveo:
Done

louis-bompart:
Why not using TypeScript tests?
https://docs.cypress.io/guides/tooling/typescript-support.html#Configure-tsconfig-json
https://github.com/bahmutov/add-typescript-to-cypress

It could allow us to refer to some code from our src easily (Like the classnames)

jeremierobert-coveo:
Here some typescript sprinkle

README.md
```diff
@@ -37,11 +37,18 @@ The code is written in [typescript](http://www.typescriptlang.org/) and compiled
...
 
 -   `npm run test` will execute the tests one time and give you the report
 -   `npm run watchTest` will watch changes and reexecute the tests and coverage when saving a file.
+
+## Running functional tests
+
+1. `npm run dev` will start the developer server.
+2. `npx cypress run` will run the cypress functional tests suite.
+
+To add a new test suite simply add a new spec file in `./cypress/integration/` folder.
```
nathanlb: "in **the** `./cypress/integration/` folder." would technically be better :sorry:

pages/attached_result.html
```diff
@@ -38,16 +38,13 @@
                         getValues: function() {
                             return attachedIds;
                         }
-                    },
```
louis-bompart: Why is it modified?

jeremierobert-coveo: I removed the stuff that was there to test user actions

jeremierobert-coveo: I removed the stuff that was there to test user actions

PR #63, opened 2020-03-06
feat(SFINT-2829): Create a Copy to clipboard component

jeremierobert-coveo:
# Options
```typescript
    /**
     * The icon that the ResultAction will display.
     * If text is provided, the button will contain that text.
     * If the HTML of an SVG image is provided, that image will be displayed in the button.
     */
    icon?: string;

    /**
     * The tooltip that displays on hovering the ResultAction.
     */
    tooltip?: string;

    /**
     * Specifies the template that will be used for the copy to clipboard.
     *
     * Default value is `${title}\n${ClickUri}`.
     */
    template?: string;
```

# UA event
```
type: Click
cause: copyToClipboard
```

# Video
https://share.vidyard.com/watch/K5stZRJGtgVgewquEkRxHw?


nathanlb:
Can we see a paste? :D

jeremierobert-coveo:
Look at the video

jeremierobert-coveo:
Oh the video dos not show my notepad =="

src/components/CopyPaste/CopyPaste.ts
```diff
@@ -0,0 +1,95 @@
...
+        });
+    }
+
+    protected doAction() {
+        this.usageAnalytics.logClickEvent({ name: 'copyToClipboard', type: 'Click' }, {}, this.result, this.element);
+        this.copyToClipboard(StringUtils.buildStringTemplateFromResult(this.options.template, this.result));
+    }
+
+    /**
+     * Sadly that's the only way of doing in in IE11 and in lockerservice.
```
lbergeron: This comment could be removed. The `copyToClipboardIE11` function name is clear enough.

src/components/CopyPaste/CopyPaste.ts
```diff
@@ -0,0 +1,95 @@
...
+     */
```
louis-bompart: Which part is specific to LockerService?
How would it be if it wasn't for LockerService?

My point is that we should not have LockerService/Salesforce specific code in this repo.

src/components/CopyPaste/Strings.ts
```diff
@@ -0,0 +1,6 @@
+import { Translation, Language } from '../../utils/translation';
+
+Translation.register(Language.English, {
+    CopyPaste_copy: 'Copy',
+    CopyPaste_copied: 'Copied !'
```
louis-bompart:
```suggestion
    CopyPaste_copied: 'Copied!'
```

https://www.btb.termiumplus.gc.ca/tcdnstyl-chap?lang=eng&lettr=chapsect7&info0=7

lbergeron: Remove whitespace before the exclamation mark.

src/sass/ResultActionMenu.scss
```diff
@@ -0,0 +1,152 @@
...
+// Fix for quickview tooltip
+.CoveoQuickview.coveo-result-actions-menu-menu-item {
+    @extend %not-selectable;
+
+    > div {
+        height: $menu-height;
+        width: $menu-height;
+    }
+
+    // Kill poper.js
```
louis-bompart:
```suggestion
    // Kill popper.js
```

src/sass/ResultActionMenu.scss
```diff
@@ -0,0 +1,152 @@
+@import './Variables.scss';
```
louis-bompart: Why is that suddenly included? It seems out of scope

tests/components/CopyPaste/CopyPaste.spec.ts
```diff
@@ -0,0 +1,68 @@
+import { IQueryResult, $$, l } from 'coveo-search-ui';
+import { createSandbox, SinonSandbox, SinonStub } from 'sinon';
+import { Mock, Fake } from 'coveo-search-ui-tests';
+import { CopyPaste } from '../../../src/components/CopyPaste/CopyPaste';
+
+describe('ResultAction', () => {
```
louis-bompart:
```suggestion
describe('CopyPaste ResultAction', () => {
```

src/components/CopyPaste/CopyPaste.ts
jeremierobert-coveo: There is no locker service specific code.

src/sass/ResultActionMenu.scss
jeremierobert-coveo: Because the result action css was broken.

src/components/ResultAction/ResultAction.ts
```diff
@@ -44,6 +42,10 @@ export abstract class ResultAction extends Component {
         tooltip: ComponentOptions.buildStringOption()
     };
 
+    public tooltip: HTMLElement;
```
lbergeron: We shouldn't expose the tooltip element. It is breaking component encapsulation.

src/sass/ResultActionMenu.scss
louis-bompart: That's seems like another issue then, please split :D

src/components/CopyPaste/CopyPaste.ts
louis-bompart: Then the comment needs to be updated

src/sass/ResultActionMenu.scss
jeremierobert-coveo: No that css is require for Copy Paste.

src/sass/ResultActionMenu.scss
```diff
@@ -0,0 +1,150 @@
...
+// Fix for quickview tooltip
+.CoveoQuickview.coveo-result-actions-menu-menu-item {
+    @extend %not-selectable;
+
+    > div {
+        height: $menu-height;
+        width: $menu-height;
+    }
+
+    // Kill poper.js
```
lbergeron: Can you use `// Override poper.js` instead? It would sound less radical ;)

src/sass/ResultActionMenu.scss
louis-bompart: Explain please, I don't understand why a new ResultAction requires changes in the menu, changes in the QuickviewAction etc.
I'm still not convinced that this is in the scope.

src/components/CopyPaste/CopyPaste.ts
jeremierobert-coveo: Why ?

src/sass/ResultActionMenu.scss
jeremierobert-coveo: Because of the existing css result actions look out of place when compared with quickview result actions.

tests/components/CopyPaste/CopyPaste.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+        element = document.createElement('div');
+        document.body.append(element);
+        testOptions = new Mock.AdvancedComponentSetupOptions(element);
+
+        result.title = 'some title';
+        result.clickUri = 'some uri';
+
+        componentSetup = Mock.advancedResultComponentSetup(CopyPaste, result, testOptions);
+        testComponent = componentSetup.cmp;
+        writeTextStub = sandbox.stub(navigator.clipboard, 'writeText').resolves();
```
lbergeron: This does not test the IE11 code path. It would be nice to have a test case that covers IE11.

src/components/CopyPaste/CopyPaste.ts
louis-bompart: It's literally written **lockerservice** in the comment, if there's no part that is specific in lockerservice, then it's only a IE11 issue and thus the `and in lockerservice` is not needed

src/sass/ResultActionMenu.scss
louis-bompart: And that's another issue :)

src/components/CopyPaste/CopyPaste.ts
jeremierobert-coveo: It's also a lockerservice issue because the Clipboard AP is not available in locker service as in IE11 end of story and locker service needs to be mention because if we remove support for IE11 we still need this code for locker service.

src/sass/ResultActionMenu.scss
jeremierobert-coveo: So you want me to ship an half feature instead of a complete one ?

src/components/CopyPaste/CopyPaste.ts
mikegron: cause it mentions lockerservice? 📦

tests/components/CopyPaste/CopyPaste.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+    describe('on click', () => {
+        it('should paste the result in the clipboard', async () => {
+            testComponent.element.click();
+            expect(writeTextStub.called).toBeTrue();
+        });
+    });
+
+    describe('on mouseleave', () => {
+        it('should change the tooltip text for "Copy"', () => {
+            testComponent.tooltip.innerText = 'patate';
```
lbergeron: I think you souldn't expose `.tooltip`. I guess you should validate this behavior using a functional test instead. If you prefer to use a unit test, maybe we could expose an event instead.

src/components/CopyPaste/CopyPaste.ts
jeremierobert-coveo: @mikegron you respond at which question ?

src/components/CopyPaste/CopyPaste.ts
```diff
@@ -0,0 +1,95 @@
...
+
+    /**
+     * Specifies the template that will be used for the copy to clipboard.
+     *
+     * Default value is `${title}\n${ClickUri}`.
+     */
+    template?: string;
+}
+
+export class CopyPaste extends ResultAction {
```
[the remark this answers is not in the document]
jeremierobert-coveo: ClickUri is just an option you can change this component to copy the excert to clipboard so it's not bound to a field

src/components/CopyPaste/CopyPaste.ts
louis-bompart: So, the clipboard API does not work in LockerService even in Chrome ?

tests/components/CopyPaste/CopyPaste.spec.ts
jeremierobert-coveo: Sadly you can't. (Trust me I try)

tests/components/CopyPaste/CopyPaste.spec.ts
jeremierobert-coveo: You can't use Functionnal tests because of some issues in cypress regarding clipboard and hovering

src/components/CopyPaste/CopyPaste.ts
jeremierobert-coveo: Exacly

jeremierobert-coveo: image

src/sass/ResultActionMenu.scss
louis-bompart: I want to ship the strict minimum for the feature, this doesn't seems like it.
Can you indicate in this file which CSS rule concerns directly and exclusively the CopyAndPaste component?

src/components/CopyPaste/CopyPaste.ts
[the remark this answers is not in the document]
lbergeron: I see your point. I think `CopyResult` would be a better name name. In my opinion, `Paste` is irrelevant as the component only copies data. I pastes nothing.

src/components/CopyPaste/CopyPaste.ts
```diff
@@ -0,0 +1,95 @@
...
+        public element: HTMLElement,
+        public options: ICopyPasteOptions,
+        public bindings?: IResultsComponentBindings,
+        public result?: IQueryResult
+    ) {
+        super(element, ComponentOptions.initComponentOptions(element, CopyPaste, options), bindings, result);
+
+        super.init();
+
+        this.bind.on(this.element, 'mouseleave', (event: MouseEvent) => {
```
mikegron: ont peux pas set le innerText au debut ? sa me fait bizzare de setté ca dans un event mais bon

src/components/CopyPaste/CopyPaste.ts
lbergeron: Why not name the method `copyToClipboardFallback` instead? It would include IE11 and LockerService without naming them explicitely.

src/components/CopyPaste/CopyPaste.ts
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    /**
+     * Sadly that's the only way of doing in in IE11 and in lockerservice.
+     */
+    private copyToClipboard(text: string) {
+        if (navigator && navigator.clipboard) {
+            navigator.clipboard
+                .writeText(text)
+                .then(() => (this.tooltip.innerText = l('CopyPaste_copied')), err => this.logger.error('Copy to clipboard failed.', text, err));
```
mikegron: l'error devrais pas être dans un `.catch()` pour simplicité ?

src/components/CopyPaste/CopyPaste.ts
```diff
@@ -0,0 +1,95 @@
...
+        } else {
+            this.copyToClipboardIE11(text);
+        }
+    }
+
+    private copyToClipboardIE11(text: string) {
```
mikegron: mes sympathies 🥀

src/sass/ResultActionMenu.scss
mikegron: I agree with jeremie

src/sass/ResultActionMenu.scss
```diff
@@ -0,0 +1,152 @@
...
+}
+
+%select-all {
+    -webkit-user-select: all;
+    -moz-user-select: all;
+    -ms-user-select: all;
+    user-select: all;
+}
+
+.CoveoResultActionsMenu {
```
mikegron: Si je comprend bien c'est un copy-paste du scss de search ui avec des petites modifications ? C'est pas possible d'avoir juste les modifs pour pas avoir ce gros fichier?

tests/components/CopyPaste/CopyPaste.spec.ts
lbergeron: Did you try spying on `document.execCommand`? Or we could add an event like `onCopy` that would contain the copied information. This way we would be able to easily validate that the right data is sent without having to deal directly with the browser.

src/sass/ResultActionMenu.scss
jeremierobert-coveo: Pas dutout, c'est pas du jsui c'est de SFINT. Une belle mise a jour

src/components/CopyPaste/CopyPaste.ts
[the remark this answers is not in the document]
jeremierobert-coveo: What about `CopyToClipboard` ?

src/components/CopyPaste/CopyPaste.ts
jeremierobert-coveo: will change the function name for `copyToClipboardFallback` so I'll keep this comment.

src/components/CopyPaste/CopyPaste.ts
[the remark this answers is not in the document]
lbergeron: Yes I'm fine with that!

src/components/CopyPaste/CopyPaste.ts
jeremierobert-coveo: Nope tu changer le texte de copy a copied!

src/components/CopyToClipboard/CopyToClipboard.ts
```diff
@@ -0,0 +1,96 @@
...
+
+export class CopyToClipboard extends ResultAction {
+    static ID = 'CopyToClipboard';
+
+    /**
+     * The possible options for _ResultAction_.
+     * @componentOptions
+     */
+    static options: ICopyToClipboardOptions = {
+        icon: ComponentOptions.buildStringOption({ defaultValue: copyToClipboard }),
```
louis-bompart: Too much breaklines imho here. Not blocking for review tho.

src/components/CopyToClipboard/CopyToClipboard.ts
```diff
@@ -0,0 +1,96 @@
...
+    protected doAction() {
+        this.usageAnalytics.logClickEvent({ name: 'copyToClipboard', type: 'Click' }, {}, this.result, this.element);
+        this.copyToClipboard(StringUtils.buildStringTemplateFromResult(this.options.template, this.result));
+    }
+
+    private copyToClipboard(text: string) {
+        if (navigator && navigator.clipboard) {
+            navigator.clipboard
+                .writeText(text)
+                .then(() => (this.element.querySelector<HTMLElement>('.coveo-caption-for-icon').innerText = l('CopyToClipboard_copied')))
```
louis-bompart: async/await instead?

src/components/ResultAction/ResultAction.ts
```diff
@@ -44,6 +42,8 @@ export abstract class ResultAction extends Component {
         tooltip: ComponentOptions.buildStringOption()
     };
 
+    private isInitialized = false;
```
louis-bompart: Mhh, revert the change on this file I'd say?

jeremierobert-coveo:
All comments resolve and change implemented

tests/components/UserActions/QueryList.spec.ts
```diff
@@ -269,30 +269,6 @@ describe('QueryList', () => {
             });
         });
 
-        it('should log a search event with "userActionsSubmit" as search cause', () => {
```
lbergeron: Why are you removing this test?

jeremierobert-coveo: usageAnalytics is no more mockable by updating the jsui

lbergeron: Upgrading from `2.0.342` to `2.8521.7` is a pretty big leap forward indeed. Ok then.

PR #66, opened 2020-03-11
SFINT-3009 Fix an issue where the text is not changed when clicked

jeremierobert-coveo:
* Added UTs.
* Added an option to set the tooltip success label.
* Ensure the success label is shown on success.

(This PR is therefore called shame)

### DIFF
--- a/src/components/CopyToClipboard/CopyToClipboard.ts
+++ b/src/components/CopyToClipboard/CopyToClipboard.ts
@@ -17,7 +17,7 @@ export interface ICopyToClipboardOptions extends IResultActionOptions {
     tooltip?: string;
 
     /**
-     * The tooltip that displays when the action succeed.
+     * The tooltip that displays when the action succeeds.
      */
     successTooltip?: string;
 
--- a/src/components/CopyToClipboard/CopyToClipboard.ts
+++ b/src/components/CopyToClipboard/CopyToClipboard.ts
@@ -17,6 +17,11 @@ export interface ICopyToClipboardOptions extends IResultActionOptions {
     tooltip?: string;
 
     /**
+     * The tooltip that displays when the action succeed.
+     */
+    successTooltip?: string;
+
+    /**
      * Specifies the template that will be used for the copy to clipboard.
      *
      * Default value is `${title}\n${ClickUri}`.
@@ -35,6 +40,8 @@ export class CopyToClipboard extends ResultAction {
         icon: ComponentOptions.buildStringOption({ defaultValue: copy }),
 
         tooltip: ComponentOptions.buildCustomOption(tooltip => tooltip, { defaultFunction: () => l('CopyToClipboard_copy') }),
+
+        successTooltip: ComponentOptions.buildCustomOption(tooltip => tooltip, { defaultFunction: () => l('CopyToClipboard_copied') }),
 
         template: ComponentOptions.buildStringOption({ defaultValue: '${title}\n${clickUri}' })
     };
@@ -58,7 +65,7 @@ export class CopyToClipboard extends ResultAction {
 
         this.bind.on(this.element, 'mouseleave', (event: MouseEvent) => {
             if (event.target == this.element) {
-                this.element.querySelector<HTMLElement>('.coveo-caption-for-icon').innerText = l('CopyToClipboard_copy');
+                this.setToolipText(this.options.tooltip);
             }
         });
     }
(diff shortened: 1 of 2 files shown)


src/components/CopyToClipboard/CopyToClipboard.ts
```diff
@@ -72,14 +79,21 @@ export class CopyToClipboard extends ResultAction {
...
                 await navigator.clipboard.writeText(text);
-                this.element.querySelector<HTMLElement>('.coveo-caption-for-icon').innerText = l('CopyToClipboard_copied');
             } catch (err) {
                 this.logger.error('Copy to clipboard failed.', text, err);
                 this.copyToClipboardFallback(text);
             }
         } else {
             this.copyToClipboardFallback(text);
         }
+        this.setToolipText(this.options.successTooltip);
```
louis-bompart: Would remove the parameter, it's stored in `this.options`

src/components/CopyToClipboard/CopyToClipboard.ts
```diff
@@ -72,14 +79,21 @@ export class CopyToClipboard extends ResultAction {
...
+    }
+
+    private setToolipText(text: string) {
+        const tooltipElement = this.element.querySelector<HTMLElement>('.coveo-caption-for-icon');
+        if (tooltipElement && text) {
```
louis-bompart: What should be the behavior if I set `this.options.successTooltip=''` ?

src/components/CopyToClipboard/CopyToClipboard.ts
```diff
@@ -16,6 +16,11 @@ export interface ICopyToClipboardOptions extends IResultActionOptions {
      */
     tooltip?: string;
 
+    /**
+     * The tooltip that displays when the action succeed.
```
lbergeron: `succeed.` --> `succeeds`

src/components/CopyToClipboard/CopyToClipboard.ts
mikegron: Je dirais de ne pas changer le texte

jeremierobert-coveo: Exact

tests/components/CopyPaste/CopyPaste.spec.ts
```diff
@@ -62,6 +70,11 @@ describe('CopyToClipboard ResultAction', () => {
                 testComponent.element.click();
                 expect(execCommandStub.called).toBeTrue();
             });
+
+            it('should change the tooltip text for "Copied!"', async () => {
```
lbergeron: You should add tests to validate tooltip text when using custom `options.tooltip` and `options.successTooltip`.

src/components/CopyToClipboard/CopyToClipboard.ts
louis-bompart: We need a test for that then :D

src/components/CopyToClipboard/CopyToClipboard.ts
jeremierobert-coveo: Nope the function is use twice with different param

tests/components/CopyPaste/CopyPaste.spec.ts
```diff
@@ -71,17 +87,59 @@ describe('CopyToClipboard ResultAction', () => {
...
+            it('should use the provided successTooltip', async () => {
+                const tooltipText = 'some success tooltip';
+                buildResultAction({ successTooltip: tooltipText });
+                testComponent.element.click();
+
+                await Promise.resolve();
+                expect(testComponent.element.querySelector<HTMLElement>('.coveo-caption-for-icon').innerText).toEqual(tooltipText);
+            });
+
+            it('should not change the tooltip if the successTooltip is empty-like', async () => {
```
lbergeron: Why not `...if the successTooltip is empty.`? Using `empty-like` hints me that you should have tested `undefined` as well.

jeremierobert-coveo: null undefined or '' is the same

lbergeron: I know but your test is checking only one of those 3 cases right now. The best would be to validate those 3 cases. If we stick to validating only the empty string, I would rename the test case to make it clear.

jeremierobert-coveo: oki

PR #68, opened 2020-03-17
fix(SFINT-3034): Make the tooltip flashing more reliable

jeremierobert-coveo:
See here the new behavior https://share.vidyard.com/watch/27BHA6A8FiegmLnT8mcR1v?

### DIFF
--- a/src/components/CopyToClipboard/CopyToClipboard.ts
+++ b/src/components/CopyToClipboard/CopyToClipboard.ts
@@ -62,12 +62,6 @@ export class CopyToClipboard extends ResultAction {
         super(element, ComponentOptions.initComponentOptions(element, CopyToClipboard, options), bindings, result);
 
         super.init();
-
-        this.bind.on(this.element, 'mouseleave', (event: MouseEvent) => {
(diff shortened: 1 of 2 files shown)


src/components/CopyToClipboard/CopyToClipboard.ts
```diff
@@ -87,6 +81,11 @@ export class CopyToClipboard extends ResultAction {
             this.copyToClipboardFallback(text);
         }
         this.setToolipText(this.options.successTooltip);
+        this.refreshTooltip();
+    }
+
+    private refreshTooltip() {
+        setTimeout(() => this.setToolipText(this.options.tooltip), 500);
```
tedre191: Can you explain why this is required now ? What was the issue with the mouseleave ?

jeremierobert-coveo: Yeah the issue with mouse leave is that the event is not consistent enough.
For instance if I pass my mouse through the tooltip no mouseleave will be trigger on the parent element.

PR #70, opened 2020-04-28
fix(SFINT-3128): No const in es5

jeremierobert-coveo:
We use es6 module with es5 js.
Const are not part of es5.


louis-bompart:
Ain't the issue that we're using es6 modules with es5?
Shouldn't we support es6 modules with es6 or higher?

mikegron:
Sa brisait qqchose?

PR #71, opened 2020-04-30
Fix/SFINT-3128 es6 is now 100% es6

jeremierobert-coveo:
Everything is in the title.

### DIFF
--- a/config/webpack.config.js
+++ b/config/webpack.config.js
@@ -29,7 +29,10 @@ module.exports = {
                 test: /\.ts$/,
                 loader: 'ts-loader',
                 options: {
-                    configFile: path.resolve('./config/tsconfig.json')
+                    configFile: path.resolve('./config/tsconfig.json'),
+                    compilerOptions: {
+                        target: 'es5'
+                    }
                 }
             },
             {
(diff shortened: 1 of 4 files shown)


jeremierobert-coveo:
C'est pas sense toucher la partie commonjs 😛

config/webpack.config.js
```diff
@@ -29,7 +29,10 @@ module.exports = {
                 test: /\.ts$/,
                 loader: 'ts-loader',
                 options: {
-                    configFile: path.resolve('./config/tsconfig.json')
+                    configFile: path.resolve('./config/tsconfig.json'),
+                    compilerOptions: {
+                        target: 'es5'
```
tedre191, line 34: Not sure why you changed the tsconfig.json to use ES6, only to force es5 here ?

jeremierobert-coveo: The es6 build will build es6 instead of es5

src/utils/icons.ts
```diff
@@ -1,5 +1,3 @@
-/// <reference path='../typings/custom.d.ts'/>
```
louis-bompart, line 1: Pretty sure typescript was freaking out last time I tried 🤔

PR #73, opened 2020-05-27
[feat/SFINT-3189] Added date to user activity timestamp

nathanlb:
- Added formatDateAndTime utility function to return full formatted date and time
- Added this new date time format to the user activity timestamp.
- Moved the timestamp beside the activity title
- Added and modified UT accordingly

Screen Shot 2020-05-29 at 11 34 08 AM
Screen Shot 2020-05-29 at 11 33 22 AM

### DIFF
--- a/src/components/UserActions/UserActivity.ts
+++ b/src/components/UserActions/UserActivity.ts
@@ -38,7 +38,10 @@ const MAIN_CLASS = 'coveo-user-activity';
 const CELL_CLASS = 'coveo-cell';
 const TITLE_CLASS = 'coveo-title';
 const DATA_CLASS = 'coveo-data';
-const TIMESTAMP_CLASS = 'coveo-footer';
+const ORIGIN_CLASS = 'coveo-footer';
+const ACTIVITY_TITLE_SECTION = 'coveo-activity-title-section';
+const ACTIVITY_TITLE_CLASS = 'coveo-activity-title';
+const ACTIVIY_TIMESTAMP_CLASS = 'coveo-activity-timestamp';
 const HEADER_CLASS = 'coveo-header';
 const ACTIVITY_CLASS = 'coveo-activity';
 
@@ -238,11 +241,11 @@ export class UserActivity extends Component {
 
         document.createAttributeNS('svg', 'svg');
 
-        li.appendChild(this.buildTitleElement(action));
+        li.appendChild(this.buildTitleSection(action));
         if (action.document) {
             li.appendChild(dataElement);
         }
-        li.appendChild(this.buildTimestampElement(action));
+        li.appendChild(this.buildOriginElement(action));
         li.appendChild(this.buildIcon(duplicate));
 
         return li;
@@ -252,7 +255,7 @@ export class UserActivity extends Component {
         const li = document.createElement('li');
         li.classList.add(SEARCH_EVENT_CLASS);
 
-        li.appendChild(this.buildTitleElement(action));
+        li.appendChild(this.buildTitleSection(action));
 
         if (action.query) {
             const dataElement = document.createElement('div');
@@ -262,7 +265,7 @@ export class UserActivity extends Component {
             li.appendChild(dataElement);
         }
 
-        li.appendChild(this.buildTimestampElement(action));
+        li.appendChild(this.buildOriginElement(action));
         li.appendChild(this.buildIcon(search));
 
         return li;
@@ -276,9 +279,9 @@ export class UserActivity extends Component {
         dataElement.classList.add(DATA_CLASS);
         dataElement.innerText = `${action.raw.content_id_key}: ${action.raw.content_id_value}`;
 
-        li.appendChild(this.buildTitleElement(action));
+        li.appendChild(this.buildTitleSection(action));
         li.appendChild(dataElement);
-        li.appendChild(this.buildTimestampElement(action));
+        li.appendChild(this.buildOriginElement(action));
         li.appendChild(this.buildIcon(view));
 
         return li;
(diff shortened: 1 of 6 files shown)


src/utils/time.ts
```diff
@@ -2,17 +2,21 @@ import { l } from 'coveo-search-ui';
...
 /**
  * Format a date object to a date string that follow the format describe below.
- * > Ex: `Mon, Apr 29, 2019`
+ * > Ex: `Mon, 29 Apr 2019`
  * @param date The date that will be formated.
  * @returns A string formated version of the date.
  */
 export function formatDate(date: Date) {
-    return date.toLocaleDateString('default', {
+    const formattedDate = date.toLocaleDateString('default', {
```
louis-bompart, line 10: Should we allow integrator or enduser to customize the time formating?

tests/components/UserActions/UserActivity.spec.ts
```diff
@@ -5,64 +5,64 @@ import { UserActionType } from '../../../src/rest/UserProfilingEndpoint';
...
         event_value: 'Case Submit' + Math.random()
     });
 
-    const FAKE_CUSTOM_EVENT_WITHOUT_TYPE = new UserAction(UserActionType.Custom, new Date('1:00:00 AM'), {
+    const FAKE_CUSTOM_EVENT_WITHOUT_TYPE = new UserAction(UserActionType.Custom, new Date(TEST_DATE_STRING), {
         origin_level_1: 'relevant' + Math.random(),
         event_value: 'Case Submit' + Math.random()
     });
 
     const IRRELEVANT_ACTIONS = [
```
louis-bompart, line 47: Do you think a 'dumber' array (i.e. with primitive) and a `map` would be better?

src/components/UserActions/UserActivity.ts
```diff
@@ -289,37 +292,56 @@ export class UserActivity extends Component {
...
         return li;
     }
 
-    private buildTimestampElement(action: UserAction) {
+    private buildOriginElement(action: UserAction): HTMLElement {
         const el = document.createElement('div');
-        el.classList.add(TIMESTAMP_CLASS);
-        el.innerText = `${formatTime(action.timestamp)}${(action.raw.origin_level_1 && ` - ${action.raw.origin_level_1}`) || ''}`;
+        el.classList.add(ORIGIN_CLASS);
+        el.innerText = `${(action.raw.origin_level_1 && action.raw.origin_level_1) || ''}`;
```
louis-bompart, line 317: Huh? I don't understand this line.

src/components/UserActions/UserActivity.ts
```diff
@@ -289,37 +292,56 @@ export class UserActivity extends Component {
...
         return el;
     }
 
-    private buildTitleElement(action: UserAction) {
+    private buildTimestampElement(action: UserAction): HTMLElement {
+        const el = document.createElement('div');
+        el.classList.add(ACTIVIY_TIMESTAMP_CLASS);
+        el.innerText = `${formatDateAndTime(action.timestamp)}`;
```
louis-bompart, line 324: the string templating is overkill here I think.

src/utils/time.ts
```diff
@@ -2,15 +2,15 @@ import { l } from 'coveo-search-ui';
 
 /**
  * Format a date object to a date string that follow the format describe below.
- * > Ex: `Mon, Apr 29, 2019`
+ * > Ex: `Mon, 29 Apr 2019`
```
louis-bompart, line 5: Is there really a change in the end 😛 ?

src/components/UserActions/UserActivity.ts
nathanlb: me neither that's just the way it was :P I'll try removing it and we'll see if the sky falls.

nathanlb: The tests pass just fine without it.

louis-bompart: @jeremierobert-coveo you worked first on this component iirc, can you check wassup?

mikegron:
Can you check if it looks ok if the width of the whole user actions div is 300px ? that's the sidebar width in servicenow at least :D

louis-bompart:
The minimal width of a SearchInterface is 300px according to the CSS of the coveo/search-ui. So indeed, that's a good criteria and the feature should comply with it :ok_hand:

src/components/UserActions/UserActivity.ts
jeremierobert-coveo: This was rendering the time. Can you check that the now string is something like `<My Datetime> - <My origin lvl 1>`

nathanlb:
@mikegron 300px is quite small so there's some overflow as you can see.
Screen Shot 2020-05-29 at 8 50 17 AM
If this is inadequate maybe an option allowing you to decide if you'd like full date or just the time?

jeremierobert-coveo:
You can use a new line

nathanlb:
> You can use a new line

I kindof feel like that's what's happening. what do you mean?

src/components/UserActions/UserActivity.ts
nathanlb: we moved the timestamp to be on the right, inline with the title. So now it just renders the origin level. What we're curious about is this: 
`action.raw.origin_level_1 && action.raw.origin_level_1 || ''`

louis-bompart:
Not fond of yet another newline, this'll be quite ugly I think.

Maybe the dateformat should be responsive.
Like, when width < Narrow_Screen, use yyyy/mm/dd - hh:mm (24hrs)

mikegron:
> Not fond of yet another newline, this'll be quite ugly I think.
> 
> Maybe the dateformat should be responsive.
> Like, when width < Narrow_Screen, use yyyy/mm/dd - hh:mm (24hrs)

I agree we should not have a newline.. Maybe we could drop the day of the week and use 24h without seconds like Louis said so something like `Jan 22, 2020 15:49` ?
We could probably also drop the Year maybe, since user actions are normally very close to "real time"! So `Jan 22, 15:49` ?

nathanlb:
> > Not fond of yet another newline, this'll be quite ugly I think.
> > Maybe the dateformat should be responsive.
> > Like, when width < Narrow_Screen, use yyyy/mm/dd - hh:mm (24hrs)
> 
> I agree we should not have a newline.. Maybe we could drop the day of the week and use 24h without seconds like Louis said so something like `Jan 22, 2020 15:49` ?
> We could probably also drop the Year maybe, since user actions are normally very close to "real time"! So `Jan 22, 15:49` ?

I like that. I'll make a a "short" function to display Jan 22, 15:49 if the display is too small. I can put the full date in the title attribute so it'll display on hover too.

src/components/UserActions/UserActivity.ts
```diff
@@ -289,37 +292,56 @@ export class UserActivity extends Component {
...
+        el.classList.add(ORIGIN_CLASS);
+        el.innerText = action.raw.origin_level_1 || '';
         return el;
     }
 
-    private buildTitleElement(action: UserAction) {
+    private buildTimestampElement(action: UserAction): HTMLElement {
+        const el = document.createElement('div');
+        el.classList.add(ACTIVIY_TIMESTAMP_CLASS);
+        el.innerText = this.element.clientWidth > 350 ? formatDateAndTime(action.timestamp) : formatDateAndTimeShort(action.timestamp);
```
louis-bompart, line 324: I would maybe use [offsetWidth](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/offsetWidth).
Reason: 

> The Element.clientWidth property is zero for inline elements and elements with no CSS;

https://developer.mozilla.org/en-US/docs/Web/API/Element/clientWidth

src/components/UserActions/UserActivity.ts
louis-bompart: 👌

PR #74, opened 2020-06-03
feat(SFINT-3167): Add ActionButton Component

lbergeron:
This PR adds the _ActionButton_ component. This component is meant to be a reusable button component displaying an icon and/or text with a tooltip.

The button looks like this:

action-button

Hovering the button looks like this:

action-button-icon_hover
action-button-icon-text_hover
action-button-text_hover

### DIFF
--- a/src/components/ActionButton/ActionButton.scss
+++ b/src/components/ActionButton/ActionButton.scss
@@ -6,8 +6,10 @@ $primary-color-light: #e5e5e5;
 $primary-color-dark: #4a4a4a;
 $accent-color: $calypso;
 
+$button-size: 36px;
+
 button.CoveoActionButton.coveo-actionbutton {
-    height: 36px;
+    height: $button-size;
     border: solid 2px $primary-color-light;
     border-radius: 3px;
     background-color: $primary-color-lightest;
@@ -19,12 +21,12 @@ button.CoveoActionButton.coveo-actionbutton {
 }
 
 .CoveoActionButton {
-    &:not(.coveo-actionbutton-icon) {
+    &:not(.coveo-actionbutton-icononly) {
         padding: 0 16px;
     }
 
-    &.coveo-actionbutton-icon {
-        width: 36px;
+    &.coveo-actionbutton-icononly {
+        width: $button-size;
     }
 
     .coveo-actionbutton_icon {
--- a/src/components/ActionButton/ActionButton.scss
+++ b/src/components/ActionButton/ActionButton.scss
@@ -1,7 +1,10 @@
+@import '../../sass/Variables.scss';
+
 $primary-color-lightest: #ffffff;
+$primary-color-lightest-hover: whitesmoke;
 $primary-color-light: #e5e5e5;
 $primary-color-dark: #4a4a4a;
-$accent-color: #1d4f76;
+$accent-color: $calypso;
 
 button.CoveoActionButton.coveo-actionbutton {
     height: 36px;
@@ -44,9 +47,11 @@ button.CoveoActionButton.coveo-actionbutton {
     }
 }
 
-.CoveoActionButton.coveo-actionbutton:hover {
+.CoveoActionButton.coveo-actionbutton:hover,
+.CoveoActionButton.coveo-actionbutton:active {
     & {
         color: $accent-color;
+        background-color: $primary-color-lightest-hover;
     }
 
     .coveo-actionbutton_icon svg {
--- a/src/components/ActionButton/ActionButton.scss
+++ b/src/components/ActionButton/ActionButton.scss
@@ -1,0 +1,55 @@
+$primary-color-lightest: #ffffff;
+$primary-color-light: #e5e5e5;
+$primary-color-dark: #4a4a4a;
+$accent-color: #1d4f76;
+
+button.CoveoActionButton.coveo-actionbutton {
+    height: 36px;
+    border: solid 2px $primary-color-light;
+    border-radius: 3px;
+    background-color: $primary-color-lightest;
+    color: $primary-color-dark;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    white-space: nowrap;
+}
+
+.CoveoActionButton {
+    &:not(.coveo-actionbutton-icon) {
+        padding: 0 16px;
+    }
+
+    &.coveo-actionbutton-icon {
+        width: 36px;
+    }
+
+    .coveo-actionbutton_icon {
+        line-height: 0;
+        font-size: 16px;
+
+        svg {
+            height: 1em;
+            width: 1em;
+            fill: $primary-color-dark;
+        }
+
+        + .coveo-actionbutton_title {
+            margin-left: 8px;
+        }
+    }
+
+    + .coveo-actionbutton {
+        margin-left: 5px;
+    }
+}
+
+.CoveoActionButton.coveo-actionbutton:hover {
+    & {
+        color: $accent-color;
+    }
+
+    .coveo-actionbutton_icon svg {
+        fill: $accent-color;
+    }
+}
--- a/src/components/ActionButton/ActionButton.ts
+++ b/src/components/ActionButton/ActionButton.ts
@@ -62,7 +62,16 @@ export class ActionButton extends Component {
          * <button class='CoveoActionButton' data-icon='&lt;svg width=&quot;1em&quot; height=&quot;1em&quot;&gt;...&lt;/svg&gt;'></button>
          * ```
          */
-        icon: ComponentOptions.buildStringOption()
+        icon: ComponentOptions.buildStringOption(),
+
+        /**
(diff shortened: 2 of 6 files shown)


src/components/ActionButton/ActionButton.scss
```diff
@@ -0,0 +1,60 @@
+@import '../../sass/Variables.scss';
+
+$primary-color-lightest: #ffffff;
+$primary-color-lightest-hover: whitesmoke;
+$primary-color-light: #e5e5e5;
+$primary-color-dark: #4a4a4a;
+$accent-color: $calypso;
+
+button.CoveoActionButton.coveo-actionbutton {
+    height: 36px;
```
jeremierobert-coveo, line 10: Can we have a variable for this magic 36px ? 😄

nathanlb:
Could we also see a screenshot of the button within a coveo search page? Just to get an idea of how it fits in aesthetically.

src/components/ActionButton/ActionButton.scss
```diff
@@ -0,0 +1,60 @@
...
+.CoveoActionButton {
+    &:not(.coveo-actionbutton-icon) {
+        padding: 0 16px;
+    }
+
+    &.coveo-actionbutton-icon {
+        width: 36px;
+    }
+
+    .coveo-actionbutton_icon {
```
nathanlb, line 30: underscore is probably meant to be hyphen.

src/components/ActionButton/ActionButton.scss
```diff
@@ -0,0 +1,60 @@
...
+        line-height: 0;
+        font-size: 16px;
+
+        svg {
+            height: 1em;
+            width: 1em;
+            fill: $primary-color-dark;
+        }
+
+        + .coveo-actionbutton_title {
```
nathanlb, line 40: here too... wait is this on purpose?

nathanlb:
Nifty little component here! Nice work

mikegron:
Nice! Is this supposed to be used in the Actions Menu or it's completely seperate?

src/components/ActionButton/ActionButton.scss
lbergeron: No, `coveo-actionbutton-icon` and `coveo-actionbutton_icon` do not represent the same thing. `coveo-actionbutton-icon` means that the action button is rendering in icon-only mode. The styling is a bit different in this case to ensure the button is a square and its proportions do not change depending on the icon proportions.

On the other hand `coveo-actionbutton_icon` represents the element that contains the icon part of the button. That's why we also have `coveo-actionbutton_title` to represent the title part of the button.

src/components/ActionButton/ActionButton.scss
lbergeron: Yes, see the other comment. This is inspired by the [SLDS naming convention](https://www.lightningdesignsystem.com/guidelines/markup-and-style/).

src/components/ActionButton/ActionButton.scss
nathanlb: Is it common practice to make classes distinct by swapping - for _ ? It seems like it can lead to confusion.

lbergeron:
> Nice! Is this supposed to be used in the Actions Menu or it's completely seperate?

It is separated from the result action menu. In the Salesforce integration, we intend to use it in the summary section above the result list. Will post a screenshot showing how we are using it.

src/components/ActionButton/ActionButton.scss
lbergeron: Yeah sure

src/components/ActionButton/ActionButton.scss
lbergeron: I get your concern. What would you think of `coveo-actionbutton-icononly`?
I would prefer having a single word to represent the "icon only" thing. But, at this moment, I don't know.

nathanlb: Ah i see! I kind of like the idea of using the _ to further signify a "mode" or "type". So `coveo-actionbutton_icononly ` seems good too. My only concern is when we have two classes with names that only differ by the "-" or " _".

louis-bompart: Good concerns 👍
Naming is hard 😛 
I'll stop my Haiku here. 🎉

src/components/ActionButton/ActionButton.ts
```diff
@@ -0,0 +1,91 @@
...
+    static ID = 'ActionButton';
+
+    /**
+     * The possible options for _ActionButton_.
+     * @componentOptions
+     */
+    static options: IActionButtonOptions = {
+        title: ComponentOptions.buildStringOption(),
+
+        tooltip: ComponentOptions.buildStringOption(),
```
louis-bompart, line 20: Lot of spaces here? Is that for comments? If so, where are they :D

src/components/ActionButton/ActionButton.ts
```diff
@@ -0,0 +1,91 @@
...
+    private appendTitle(): void {
+        this.element.appendChild(this.createTitleElement());
+    }
+
+    private appendTooltip(): void {
+        this.element.title = this.options.tooltip;
+    }
+}
+
+Initialization.registerAutoCreateComponent(ActionButton);
```
louis-bompart, line 91: Do we want to register it in the end?

tests/components/ActionButton/ActionButton.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+    beforeEach(() => {
+        options = {
+            title: 'a default title',
+            tooltip: 'a default tooltip',
+            icon: icons.copy
+        };
+
+        testSubject = createActionButton(options);
+
+        appendIconSpy = sandbox.spy(<any>ActionButton.prototype, 'appendIcon');
```
louis-bompart, line 28: I don't agree with this way of testing.
I think you should assert the DOM.

My rule of thumb is that if we're breaking encapsulation to test, then there's a problem

tests/components/ActionButton/ActionButton.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+
+    const createActionButton = (options: IActionButtonOptions) => {
+        const element = document.createElement('button');
+        const componentSetup = Mock.advancedComponentSetup<ActionButton>(ActionButton, new Mock.AdvancedComponentSetupOptions(element, options));
+
+        return componentSetup.cmp;
+    };
+
+    const setOption = (optionName: string, optionValue: any) => {
+        const dictOptions = options as { [key: string]: any };
```
louis-bompart, line 117: I think you should define the type on declaration, L20

src/components/ActionButton/ActionButton.scss
```diff
@@ -0,0 +1,62 @@
+@import '../../sass/Variables.scss';
+
+$primary-color-lightest: #ffffff;
```
louis-bompart, line 3: that's white nah?

mikegron: did you assume my color ? :trollface:

louis-bompart: Oh, no.

tests/components/ActionButton/ActionButton.spec.ts
louis-bompart: I agree that with complex components it's not necessarily a good idea to assert on the DOM, but when I see what you're trying to assert (i.e. Did I append the icon?) the DOM assertion make more sense.

(I was more frowning about the encapsulation break tbh)

lbergeron: I get your point. You are right. There is no need to break encapsulation to validate the expectations here.

tests/components/ActionButton/ActionButton.spec.ts
lbergeron: I don't think so. The purpose of this method is solely to be able to dynamically assign option values for testing. The type `{ [key: string]: any }` is expected to be used only in the scope of this particular method.

src/components/ActionButton/ActionButton.ts
```diff
@@ -62,7 +62,16 @@ export class ActionButton extends Component {
...
+        icon: ComponentOptions.buildStringOption(),
+
+        /**
+         * Specifies the handler called when the button is clicked.
+         *
+         * Default is `null`.
+         *
+         * This option must be set in JavaScript when initializing the component.
+         */
+        click: ComponentOptions.buildCustomOption(s => null)
```
louis-bompart, line 74: It's been a few since I used buildCustomOption, but ain't there anything to precise that there's no default value and that it **must** be set?

lbergeron: You are perfectly right, I was not aware of `IComponentOptions.required`. I was able to remove the condition and warning to use that instead.

PR #76, opened 2020-06-18
feat(sfint-3272): Add ToggleActionButton component

lbergeron:
This PR adds the _ToggleActionButton_ component. As its name states, it simply is an _ActionButton_ featuring a toggle state. In addition, to keep things simple, this button supports only having an icon and tooltip. A different icon and tooltip can be displayed when the button is activated.

Using the options, the caller might attach handlers when the button is either clicked, activated, or deactivated.

### DIFF
--- a/src/components/ActionButton/ActionButton.ts
+++ b/src/components/ActionButton/ActionButton.ts
@@ -1,9 +1,9 @@
 import { Component, ComponentOptions, IResultsComponentBindings, Initialization } from 'coveo-search-ui';
 
 export interface IActionButtonOptions {
+    icon?: string;
     title?: string;
     tooltip?: string;
-    icon?: string;
     click?: () => void;
 }
 
@@ -88,6 +88,19 @@ export class ActionButton extends Component {
 
         if (this.options.click) {
             Coveo.$$(element).on('click', () => this.options.click());
+        }
+    }
+
+    public updateIcon(icon: string): void {
+        const iconElement = this.element.querySelector('.coveo-actionbutton_icon');
+        if (iconElement && icon && icon != iconElement.innerHTML) {
+            iconElement.innerHTML = icon;
+        }
+    }
+
+    public updateTooltip(tooltip: string): void {
+        if (tooltip && tooltip != this.element.title) {
+            this.element.title = tooltip;
         }
     }
 
--- a/src/components/ActionButton/ActionButton.ts
+++ b/src/components/ActionButton/ActionButton.ts
@@ -91,6 +91,10 @@ export class ActionButton extends Component {
         }
     }
 
+    /**
+     * Updates the button icon.
+     * @param icon Markup of the SVG icon to set.
+     */
     public updateIcon(icon: string): void {
         const iconElement = this.element.querySelector('.coveo-actionbutton_icon');
         if (iconElement && icon && icon != iconElement.innerHTML) {
@@ -98,6 +102,10 @@ export class ActionButton extends Component {
         }
     }
 
+    /**
+     * Updates the button tooltip.
+     * @param tooltip The tooltip to set.
+     */
     public updateTooltip(tooltip: string): void {
         if (tooltip && tooltip != this.element.title) {
             this.element.title = tooltip;
--- a/src/components/ActionButton/ToggleActionButton.ts
+++ b/src/components/ActionButton/ToggleActionButton.ts
@@ -13,6 +13,7 @@ export interface IToggleActionButtonOptions {
 
 export class ToggleActionButton extends Component {
     static ID = 'ToggleActionButton';
+    static ACTIVATED_CLASS_NAME = 'coveo-toggleactionbutton-activated';
 
     static options: IToggleActionButtonOptions = {
         /**
@@ -76,16 +77,35 @@ export class ToggleActionButton extends Component {
          */
         tooltip: ComponentOptions.buildStringOption(),
 
+        /**
+         * Specifies the handler called when the button is clicked.
+         *
+         * Default is `null`.
+         *
+         * This option is set in JavaScript when initializing the component.
+         */
         click: ComponentOptions.buildCustomOption(s => null),
 
+        /**
+         * Specifies the handler called when the button is activated.
+         *
+         * Default is `null`.
+         *
+         * This option is set in JavaScript when initializing the component.
+         */
         activate: ComponentOptions.buildCustomOption(s => null),
 
+        /**
+         * Specifies the handler called when the button is deactivated.
+         *
+         * Default is `null`.
+         *
+         * This option is set in JavaScript when initializing the component.
+         */
         deactivate: ComponentOptions.buildCustomOption(s => null)
     };
 
-    private static ACTIVATED_CLASS_NAME = 'coveo-toggleactionbutton-activated';
-
-    private isActivated: boolean = false;
+    private _isActivated: boolean = false;
     private innerButton: ActionButton;
 
     constructor(public element: HTMLElement, public options: IToggleActionButtonOptions, public bindings?: IResultsComponentBindings) {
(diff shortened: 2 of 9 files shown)


src/components/ActionButton/ActionButton.ts
```diff
@@ -91,6 +91,19 @@ export class ActionButton extends Component {
         }
     }
 
+    public updateIcon(icon: string): void {
```
lbergeron, line 94: Oops! Forgot headers for added public methods.

nathanlb:
Looks great! Just out of curiosity had you considered extending the ActionButton as an alternative? If yes then would you mind explaining why you went this route instead?

lbergeron:
> Looks great! Just out of curiosity had you considered extending the ActionButton as an alternative? If yes then would you mind explaining why you went this route instead?

Yes, and yes. By extending _ActionButton_, _ToggleActionButton_ would also have to support all possible combinations of icon/title. Composition allows to implement only what is needed right now. If we ever need to have toggle buttons with text, it would be possible to add it.

nathanlb:
> > Looks great! Just out of curiosity had you considered extending the ActionButton as an alternative? If yes then would you mind explaining why you went this route instead?
> 
> Yes, and yes. By extending _ActionButton_, _ToggleActionButton_ would also have to support all possible combinations of icon/title. Composition allows to implement only what is needed right now. If we ever need to have toggle buttons with text, it would be possible to add it.

All right thanks. That was my suspicion. Well, GG! 🏆

src/components/ActionButton/ActionButton.ts
jeremierobert-coveo: Why is this method public ?

louis-bompart: > Why is this method public ?

Because the ActionToggleButton needs to call it. (beware, we're composing, not extending, so no, protected doesn't work)

src/components/ActionButton/ToggleActionButton.ts
```diff
@@ -0,0 +1,182 @@
...
+         *
+         * Default is `null`.
+         *
+         * This option is set in JavaScript when initializing the component.
+         */
+        deactivate: ComponentOptions.buildCustomOption(s => null)
+    };
+
+    private _isActivated: boolean = false;
+    private innerButton: ActionButton;
```
louis-bompart, line 109: `innerActionButton` I'd say.
Even if that's a bit heavy handed, that leave no room for misundersanding.

src/components/ActionButton/ActionButton.ts
lbergeron: Since _ToggleActionButton_ uses the _ActionButton_ (not extending it), _ToggleActionButton_ can update the icon after the fact. It is the same idea for the tooltip.

We need to be able to use a different icon when the button is activated.

src/components/ActionButton/ToggleActionButton.ts
```diff
@@ -0,0 +1,182 @@
...
+import { ActionButton } from './ActionButton';
+
+export interface IToggleActionButtonOptions {
+    activatedIcon: string;
+    activatedTooltip: string;
+    icon: string;
+    tooltip: string;
+    click?: () => void;
+    activate?: () => void;
+    deactivate?: () => void;
```
louis-bompart, line 11: I would name them `onClick`, `onActivate`, `onDeactivate`

src/components/ActionButton/ActionButton.ts
jeremierobert-coveo: Indeed, I'm always warry of new public method 😛

src/components/ActionButton/ToggleActionButton.ts
```diff
@@ -0,0 +1,182 @@
...
+            bindings
+        );
+
+        this.updateButton();
+    }
+
+    private updateButton() {
+        if (this._isActivated) {
+            this.element.classList.add(ToggleActionButton.ACTIVATED_CLASS_NAME);
+            this.element.setAttribute('aria-pressed', 'true');
```
louis-bompart, line 168: TIL: If you want a button to be considered a toggle for ARIA, the attribute need to be set no matter what.
You cannot use the shortcut of just having the attribute.

src/components/ActionButton/ToggleActionButton.ts
```diff
@@ -0,0 +1,182 @@
...
+         * <svg width="1em" height="1em">...</svg>
+         * ```
+         *
+         * The attribute would be set like this:
+         *
+         * ```html
+         * <button class='CoveoToggleActionButton' data-icon='&lt;svg width=&quot;1em&quot; height=&quot;1em&quot;&gt;...&lt;/svg&gt;'></button>
+         * ```
+         */
+        icon: ComponentOptions.buildStringOption(),
```
louis-bompart, line 67: I would use `deactivatedIcon` for the name of the variable, to have the code be semantically coherent with the activated one.

tests/components/ActionButton/ToggleActionButton.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+    let updateTooltipSpy: SinonSpy;
+
+    beforeAll(() => {
+        sandbox = createSandbox();
+
+        clickSpy = spy();
+        activateSpy = spy();
+        deactivateSpy = spy();
+        updateIconSpy = spy(<any>ActionButton.prototype, 'updateIcon');
+        updateTooltipSpy = spy(<any>ActionButton.prototype, 'updateTooltip');
```
louis-bompart, line 25: Why are you not using the sandbox?

tests/components/ActionButton/ToggleActionButton.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+
+        it('should update button with activated tooltip', () => {
+            expect(updateTooltipSpy.calledWith(options.activatedTooltip)).toBeTrue();
+        });
+    });
+
+    describe('deactivating the button', () => {
+        beforeEach(() => {
+            testSubject.setActivated(true);
+            [clickSpy, activateSpy, deactivateSpy, updateIconSpy, updateTooltipSpy].forEach(s => s.resetHistory());
```
louis-bompart, line 101: Why no sandbox.reset() ?

mikegron:
Très nice! J'aime bien ces petits components la :P 
Question comme ca, est-ce qu'ont pourrais utiliser ce component la dans le component de AttachResult ou même le remplacer complètement, vue que lui aussi a un state de toggle avec une checkbox, _mais_ se trouve dans le ResultActionMenu ?
Le seul truc additionnel dans AttachResult c'est l'option pour aller chercher le state initial je crois bien.

src/components/ActionButton/ToggleActionButton.ts
lbergeron: I get your point, but I think the names are more coherent with the other components this way.

src/components/ActionButton/ToggleActionButton.ts
lbergeron: Yes, but the attribute is always set. The constructor calls `createInnerButton`, which in turn calls `updateButton`. So, the `aria-pressed` attribute is set right from the first rendering.

tests/components/ActionButton/ToggleActionButton.spec.ts
lbergeron: You're totally right. It should use `sandbox` there.

tests/components/ActionButton/ToggleActionButton.spec.ts
lbergeron: You're right, it would do the same job.

src/components/ActionButton/ToggleActionButton.ts
lbergeron: Yes, it makes sense.

lbergeron:
> Très nice! J'aime bien ces petits components la :P
> Question comme ca, est-ce qu'ont pourrais utiliser ce component la dans le component de AttachResult ou même le remplacer complètement, vue que lui aussi a un state de toggle avec une checkbox, _mais_ se trouve dans le ResultActionMenu ?
> Le seul truc additionnel dans AttachResult c'est l'option pour aller chercher le state initial et les custom Events de attach/detach je crois bien.

C'est une bonne question, je n'y avais pas songé, mais c'est vrai que les boutons ont un comportement qui se ressemble beaucoup. Je ne sais pas si le changement pourrait se faire sans briser la compatibilité par contre.

PR #77, opened 2020-07-03
Added case and article ids to AttachResult custom events

mikegron:
Small addition to the AttachResult component to specify the caseId and the field where the articleId can be obtained. For ServiceNow, the articleId is pretty much always the permanentID and the caseID is static on the page. Tested using the local test page.

- Added case and article ids to AttachResult custom events
- Added click event when attaching in AttachResult

SNOW-364

### DIFF
--- a/src/components/AttachResult/AttachResult.ts
+++ b/src/components/AttachResult/AttachResult.ts
@@ -23,6 +23,10 @@ export interface IAttachResultOptions {
     attachCaption?: string;
     /** Specifies the tooltip displayed when the result is already attached. */
     detachCaption?: string;
+    /** The field of the query result which represents the Article ID in the index, used for Usage Analytics purposes. */
+    articleIdField?: string;
+    /** Id of the record where the search is currently displayed, if any. */
+    caseId?: string;
     /** The function that is called when the user wants to attach a result. */
     attach?: (queryResult: IQueryResult) => Promise<void>;
     /** The function called when the user wants to un-link a result. */
@@ -50,6 +54,10 @@ export class AttachResult extends Component {
         detachCaption: ComponentOptions.buildStringOption({
             defaultValue: l(`${AttachResult.ID}_Detach`)
         }),
+        articleIdField: ComponentOptions.buildStringOption({
+            defaultValue: 'permanentid'
+        }),
+        caseId: ComponentOptions.buildStringOption(),
         attach: ComponentOptions.buildCustomOption(
             name => (result: IQueryResult) =>
                 new Promise<void>((resolve, reject) => {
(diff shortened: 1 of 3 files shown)


src/components/AttachResult/AttachResult.ts
```diff
@@ -50,6 +54,10 @@ export class AttachResult extends Component {
         detachCaption: ComponentOptions.buildStringOption({
             defaultValue: l(`${AttachResult.ID}_Detach`)
         }),
+        articleIdField: ComponentOptions.buildStringOption({
+            defaultValue: 'permanentid'
```
louis-bompart, line 58: Why should it take this default value?

src/components/AttachResult/AttachResult.ts
```diff
@@ -194,8 +203,11 @@ export class AttachResult extends Component {
         let customData: IAnalyticsCaseAttachMeta = {
             resultUriHash: this.queryResult.raw.urihash,
             author: this.queryResult.raw.author,
-            articleID: null,
-            caseID: null
+            articleID:
+                this.options.articleIdField && this.queryResult.raw[this.options.articleIdField]
+                    ? this.queryResult.raw[this.options.articleIdField]
+                    : null,
```
louis-bompart, line 209: Can be simplified like so:
`articleID: this.queryResult.raw[this.options.articleIdField] ?? null`
I think.

src/components/AttachResult/AttachResult.ts
```diff
@@ -194,8 +203,11 @@ export class AttachResult extends Component {
...
+            caseID: this.options.caseId
```
lbergeron, line 210: Is there are reason why you use a different casing here compared to the option name? I would use the same casing everywhere.

`caseID` -> `caseId`
`articleID` -> `articleId`

mikegron: je travaille sur les tests! :D
le casing cétait pour que sa fit avec les alias dans les dom attributes, sinon sa ferais `data-case-i-d`

src/components/AttachResult/AttachResult.ts
mikegron: that's the default ML uses to compute things so it's the most relevant field normally (ml-wise)

src/components/AttachResult/AttachResult.ts
lbergeron: Pour le casing, j'irais comme ceci. C'est bien ce que tu avais en tête?

    caseId: this.options.caseId

lbergeron: Je viens de voir que `articleID` et `caseID` étaient déjà définies. Laisses le casing tel qu'il est alors. J'étais sur l'impression que c'était de nouveaux attributs.

PR #78, opened 2020-07-03
feat(SFINT-3300): Improve ToggleActionButton option names

lbergeron:
Some of the _ToggleActionButton_ options were confusing because their value was kind of the inverse of the option name. For example, the `activatedTooltip` option could have a value like `Deactivate feature xyz`.

These options were renamed like this:

    activatedTooltip   --> deactivateTooltip
    activatedIcon      --> deactivateIcon
    deactivatedTooltip --> activateTooltip
    deactivatedIcon    --> activateIcon

This way the option values would read like this:

    activateTooltip: 'Activate feature xyz'
    deactivateTooltip: 'Deactivate feature xyz'

### DIFF
--- a/src/components/ActionButton/ToggleActionButton.ts
+++ b/src/components/ActionButton/ToggleActionButton.ts
@@ -2,10 +2,10 @@ import { ComponentOptions, IResultsComponentBindings, Component, Initialization 
 import { ActionButton } from './ActionButton';
 
 export interface IToggleActionButtonOptions {
-    activatedIcon: string;
-    activatedTooltip: string;
-    deactivatedIcon: string;
-    deactivatedTooltip: string;
+    activateIcon: string;
+    activateTooltip: string;
+    deactivateIcon: string;
+    deactivateTooltip: string;
     click?: () => void;
     activate?: () => void;
     deactivate?: () => void;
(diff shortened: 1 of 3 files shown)


src/components/ActionButton/ToggleActionButton.ts
```diff
@@ -30,25 +31,24 @@ export class ToggleActionButton extends Component {
...
          *
          * Default is the empty string.
          *
          * ```html
-         * <button class='CoveoToggleActionButton' data-activated-tooltip='My activated button tooltip'></button>
+         * <button class='CoveoToggleActionButton' data-activate-tooltip='Activate the feature'></button>
          * ```
          */
-        activatedTooltip: ComponentOptions.buildStringOption(),
+        activateTooltip: ComponentOptions.buildStringOption({ alias: 'deactivatedTooltip' }),
```
louis-bompart, line 48: Do we really want to bother with the alias?
I'd be fine with breaking, JSUI-extension is still sub 0 so, sem-ver wise, everyone that uses it should know that it's at their own risks.

lbergeron: :D I almost sent the PR without the backward compatibility support. But, I wasn't happy about consciously introducing a breaking change that could easily be handled.

PR #79, opened 2020-07-06
chore(SFINT-3304): Add backward compatibility disclaimer

lbergeron:
This PR adds a disclaimer regarding the backward compatibility for this repository.

* Indicates the project is in its pre-release phase.
* Indicates JavaScript compatibility is not guaranteed.
* Indicates TypeScript definitions are for internal use.


lbergeron:
> Maybe we should add use at your own risk 🤔

Well "use at your own risks" sounds blunt quite blunt for me. This says the same thing:

> As new releases could impact your implementation, we strongly recommend that you validate the content of each new release before upgrading to a newer version.

PR #80, opened 2020-07-08
chore(SFINT-3308): Update prettier and ran it

louis-bompart:
TL;DR:
- `npm i prettier@latest -E -D` installed latest prettier version
- `npm run lint-fix` format all the project
- `git add . ; git commit` add every change and commit.

I choose deliberately to not do any changes to the prettier config and just follow what's recommended by prettier for the new stuff - so don't add anything in prettierrc, in a perfect world, you should not even have one-.

You can find information about the changes here:
https://prettier.io/blog/2020/03/21/2.0.0.html

### DIFF
--- a/.prettierrc.js
+++ b/.prettierrc.js
@@ -9,8 +9,8 @@ module.exports = {
         {
             files: '*.json',
             options: {
-                singleQuote: false
-            }
-        }
-    ]
+                singleQuote: false,
+            },
+        },
+    ],
 };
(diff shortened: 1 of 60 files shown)


louis-bompart:
> It looks good to me. You should update the readme as Prettier dropped the support of NodeJS < 10.
> 
> https://github.com/coveo/search-ui-extensions/blob/master/README.md#setup

Agreed, but to keep the 'linting' in its own PR, I will modify it in another PR.

PR #81, opened 2020-07-08
fixeat(SFINT-3295): Improve UserActions interactivity

louis-bompart:
# Changes
## Add useResponsiveManager option
Add a useResponsiveManager option, default to true. Previous behaviour is kept, meaning that if you have hidden, the useResponsiveManager will be inoperant because hidden does super-seed it.

It allows SFINT to use the UserActions without hooking it to the ResponsiveComponentManager, will preserving its other functionality.

## Add events for show and hide of the UserActions panel
Allow other components to be notified of when the UserActions panel is shown or hidden.

It allows SFINT to sync up the state of the ToggleActionButton with the panel when the latter get closed by something else than the button (e.g. query)

# Tests
- UT
- SAN checked in SFINT

### DIFF
--- a/src/components/UserActions/UserActions.ts
+++ b/src/components/UserActions/UserActions.ts
@@ -70,6 +70,13 @@ export interface IUserActionsOptions {
      * Default: `False`
      */
     hidden: Boolean;
+    /**
+     * Whether or not the UserAction component should use the CoveoSearchUI ResponsiveManager
+     * Inoperant if `hidden` is true.
+     *
+     * Default: `True`
+     */
+    useResponsiveManager: Boolean;
 }
 
 /**
@@ -100,6 +107,9 @@ export class UserActions extends Component {
         }),
         hidden: ComponentOptions.buildBooleanOption({
             defaultValue: false
+        }),
+        useResponsiveManager: ComponentOptions.buildBooleanOption({
+            defaultValue: true
         })
     };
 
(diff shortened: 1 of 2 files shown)


src/components/UserActions/UserActions.ts
```diff
@@ -144,33 +160,45 @@ export class UserActions extends Component {
...
+
+            try {
+                const userActions = await (get(this.root, UserProfileModel) as UserProfileModel).getActions(this.options.userId);
+                if (userActions.length > 0) {
+                    this.render();
+                } else {
+                    this.renderNoActions();
+                }
+            } catch (e) {
+                if (e && e.statusCode === 404) {
```
louis-bompart, line 185: Wanted to use optional chaining, was not able because of prettier, so #80 should fix it and allow latest TypeScript feature to be used.

enadeauCoveo, line 185: Why are we only checking for status code 404?

louis-bompart: I think we're still considering the others status code, but do not wish to take special action for them other than render no actions.

I'm guessing that if https://platformqa.cloud.coveo.com/docs/?api=Reveal#!/User32Profiles/rest_organizations_paramId_machinelearning_user_actions_post answer 404, it means the feature is not enabled on the client organization.

PR #82, opened 2020-07-08
Update node requirement to node 10

louis-bompart:
Self-explanatory.
Come from #80

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ This repository contains new components meant to be used in conjunction with the
 
 ## Setup
 
-Requires Node JS >= 8.0.
+Requires Node JS >= 10.13.0.
 
 1. `npm install`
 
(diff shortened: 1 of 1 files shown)


louis-bompart:
Lodash do an update,
Builds are red,
Blocked is the PR.

README.md
```diff
@@ -9,7 +9,7 @@ This repository contains new components meant to be used in conjunction with the
 
 ## Setup
 
-Requires Node JS >= 8.0.
+Requires Node JS >= 10.13.0.
```
lbergeron, line 12: Out of curiosity, why 10.13.0 specifically? The latest 10.x version is 10.21.0.

louis-bompart: Bad google cache, I was checking the snippet from Node JS. Will update. Imposing Node 12.18.2, because it's the current LTS.

I'm a fervent partisan to try to always be on the current LTS, and not the maintenance LTS, just because its a good way to prevent late update.

PR #83, opened 2020-07-15
feat(SFINT-3327): Add disable support for ActionButtons

louis-bompart:
ActionButtons can now be disabled

When an ActionButton is disabled:
- It cannot be interacted by the user (hover, focus, click)
- Its styling changes:
image

UT, SAN-Checked.

### DIFF
--- a/pages/action_button.html
+++ b/pages/action_button.html
@@ -61,5 +61,33 @@
                 <button class="CoveoActionButton" data-title="Third"></button>
             </div>
         </div>
+
+        <div id="disabled-section">
+            <h2>Disabled buttons</h2>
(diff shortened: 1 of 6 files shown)


nathanlb:
Do we want the disabled status to affect the tooltip as well? maybe add a "(disabled)" or something?

louis-bompart:
> Do we want the disabled status to affect the tooltip as well? maybe add a "(disabled)" or something?

Good point. I'm gonna refine that... Quite a bit actually.

louis-bompart:
Blowing up in scope. Will reopen as a draft.

PR #84, opened 2020-07-15
feat(SFINT-3327): Generalize state for buttons.

louis-bompart:
# TL;DR

- Refactor ToggleActionButton upon the new shiny StatefulActionButton.
- StatefulActionButton is mostly a Finite State Machine slapped on top of the ActionButton.

# Motivations

- Keep the existing button behaviour, no addition, no removal. They do the job they're made for.
- Don't mix the states between activated/deactivated and enabled/disabled.
- Reusability, you can now have as much as state as you want.

# Tests

- SAN Check
- UT:
  - Existing 🟢 -which is nice already 'cause of the refacto-
  - StatefulButton -ToDo


src/components/ActionButton/ToggleActionButton.ts
```diff
@@ -12,6 +12,46 @@ export interface IToggleActionButtonOptions {
...
+    /**
+     * Create the activated state for a given ToggleActionButton
+     * @param button {ToggleActionButton}
+     */
+    static generateActivatedStateInstance(button: ToggleActionButton): IStatefulActionButtonState {
+        const activate = button.options.activate?.bind(this);
+        const deactivate = button.options.activate?.bind(this);
+        return {
+            onStateEntry: function () {
+                this.element.classList.add(ToggleActionButton.ACTIVATED_CLASS_NAME);
```
nathanlb, line 36: I know it's a draft but i would definitely put these two functions as private members somewhere else in the class

src/components/ActionButton/StatefulActionButton.ts
```diff
@@ -0,0 +1,131 @@
...
+        this.innerActionButton = new ActionButton(element, { ...this.options.initalState, click: this.handleClick.bind(this) }, bindings);
+    }
+
+    /**
+     * Switch the state of the instance if the state and the transition between the current and new state are allowed.
+     * @param state a state to try to switch to
+     */
+    public switchTo(state: IStatefulActionButtonState) {
+        if (!this.isTransitionAllowed(state)) {
+            console.warn('This state is not allowed on this StatefulActionButton.');
```
lbergeron, line 74: If would be nice if `state` had a `name` or `key` attribute we could use to format the warning message. It would be easier to troubleshoot.

src/components/ActionButton/StatefulActionButton.ts
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    /**
+     * Check if a transition from the current state to @param state is allowed.
+     * @param state the destination of the transition
+     */
+    private isTransitionAllowed(state: IStatefulActionButtonState) {
+        return (
+            !this.options.allowedTransitions ||
+            this.options.allowedTransitions.some((transition) => transition.from === this.currentState && transition.to === state)
```
lbergeron, line 121: I am not sure about relying on object references here for the comparisons. If the concrete component is not implemented exactly as expected, the comparison will fail, even for objects that look exactly the same.

src/components/ActionButton/ToggleActionButton.ts
```diff
@@ -12,6 +12,44 @@ export interface IToggleActionButtonOptions {
 }
 
 export class ToggleActionButton extends Component {
+    /**
+     * Create the deactivated state for a given ToggleActionButton
+     * @param button {ToggleActionButton}
+     */
+    static generateDeactivatedStateInstance(button: ToggleActionButton): IStatefulActionButtonState {
```
lbergeron, line 19: Why are these methods exposed? I would make them private instance methods instead.

louis-bompart: They could, but given what'll be coming in louis-bompart/search-ui-extensions#2 I don't think it's a matter for concerns because state definitions will be split from their use cases for code clarity.

src/components/ActionButton/StatefulActionButton.ts
louis-bompart: I understand your point and hesitated with id and whatnot, but preferred to go the strict way by using object references.
My main concern was that I don't want to deal with custom equality checkers, nor to deal with the function (`onStateEntry` or `onStateExit`) checks.

src/components/ActionButton/StatefulActionButton.ts
louis-bompart: A name could be nice, I agree, if it's only used for logging purposes (i.e. it's just a name, not a stateId ;) )

lbergeron: Yes, that's what I had in mind (using an informative name). Indeed, we could not use it to compare states.

src/components/ActionButton/StatefulActionButton.ts
lbergeron: I guess it might not be a problem to compare states using a state ID after all. Theoretically, one could alter the states to mess with the button behavior. One could do the same thing against the actual code.

On the good side, comparing IDs instead of references would make it easier to use.

louis-bompart: > Theoretically, one could alter the states to mess with the button behavior. One could do the same thing against the actual code.

Of course yes, one could break the encapsulation, but then he also breaks the implicit contract that offers a strongly typed library and he alone should deal with the pandemonium that he may face.
This is why states and transitions are specified on the constructor and are not exposed anywhere else.

The 'stringent' aspect of using object reference is my goal and I want to use 'statics' or force the reuse of the same instance to ensure no cross-contamination can occur and so the user of the button is absolutely certain of what he does when he calls `switchTo`.
Also, it's neat because most IDE will allow to go back easily to the declaration of the state, allowing you to see quite easily what are the peculiarity of your state.

lbergeron: I get your point. The part that worries me is that an implementer relying on `StatefulActionButton` has to know how the states are compared internally by `StatefulActionButton` in order to correctly implement a custom button.

I think that having a map of the states and then referring to the states by IDs in the transitions would achieve the same goal. In fact, `StatefulActionButton` would perform a lookup in the map when switching from state to state. So, in the end, it would reuse the same state instances. But I think it would be less error prone for the implementer.

src/components/ActionButton/StatefulActionButton.ts
```diff
@@ -0,0 +1,139 @@
...
+
+    /**
+     * Check if a transition from the current state to @param state is allowed.
+     * @param state the destination of the transition
+     */
+    private isTransitionAllowed(state: IStatefulActionButtonState) {
+        return (
+            this.options.states.indexOf(state) > -1 &&
+            (!this.options.allowedTransitions ||
+                this.options.allowedTransitions.some((transition) => transition.from === this.currentState && transition.to === state))
```
lbergeron, line 129: As discussed offline, we can continue to compare the states by reference. But we should add more validations to ensure that `states` and `allowedTransitions` are using the exact same references. Also, it should raise an error if the implementer passes an unknown state reference to the `switchTo` method.

louis-bompart: 👍 Should be good now ;)

louis-bompart:
> The option validations are good, but I think we should raise errors instead of warnings during initialization, just to make sure mistakes in states and transitions cannot fly under the radar and cause issues later on. I would stick to warnings in the `switchTo` method though.
> 
> I would also suggest to enforce the use of state names so the warning/error messages provide more context to the implementer and helps him fix his component.

I think we should be consistent with what's being done in `ActionButton.ts`, where warnings are used.

src/components/ActionButton/StatefulActionButton.ts
```diff
@@ -0,0 +1,162 @@
+import { IResultsComponentBindings } from 'coveo-search-ui';
+import { ActionButton, IActionButtonOptions } from './ActionButton';
+
+/**
+ * Represent a state that can be used by a StatefulActionButton.
+ */
+export type IStatefulActionButtonState = IActionButtonOptions & IStatefulActionButtonStateFunctionsAndUtils;
+
+interface IStatefulActionButtonStateFunctionsAndUtils {
```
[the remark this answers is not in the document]
louis-bompart: `IActionButtonOptions` is an Union Type. I will change its name to better reflect this.

src/components/ActionButton/StatefulActionButton.ts
```diff
@@ -0,0 +1,162 @@
...
+ * An action button able to handle multiple states and their transitions.
+ */
+export class StatefulActionButton {
+    static ID = 'StatefulActionButton';
+    private currentState: IStatefulActionButtonState;
+    private innerActionButton: ActionButton;
+
+    constructor(public element: HTMLElement, public options: IStatefulActionButtonOptions, public bindings?: IResultsComponentBindings) {
+        if (!this.areOptionsValid()) {
+            return;
```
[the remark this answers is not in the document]
louis-bompart: The `ActionButton` (not stateful) is also receiving options defined by the implementer, and we do not throw errors.
I think we should be consistent on how we handle initialization parameters.
So either we should change the warning of the ActionButton to errors - but one could say that it's a breaking change :P - or we keep warnings for the StatefulActionButton.

src/components/ActionButton/StatefulActionButton.ts
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    /**
+     * Switch the state of the instance if the state and the transition between the current and new state are allowed.
+     * @param state a state to try to switch to
+     */
+    public switchTo(state: IStatefulActionButtonState) {
+        if (!this.isTransitionAllowed(state)) {
+            console.warn(
+                `State '${state.name}' is not allowed on this StatefulActionButton.\nEnsure to use the object references used at the instantiation`
```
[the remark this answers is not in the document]
louis-bompart: I thought about it and, tbh, was lazy to change it. But agreed, will do.

tests/components/ActionButton/StatefulActionButton.spec.ts
```diff
@@ -0,0 +1,207 @@
...
+                        initalState: initialState,
+                    });
+                },
+            },
+        ].forEach((testCase) => {
+            describe(testCase.describe, () => {
+                beforeEach(testCase.beforeEach);
+
+                it(`should hide the button, log an appropriate warning and not complete the initialization`, () => {
+                    spyConsoleWarn.calledWithExactly('The stateful action button cannot render if no states are defined.');
```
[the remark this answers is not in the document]
louis-bompart: I adressed error v.s. warning on my general comment and above.

tests/components/ActionButton/StatefulActionButton.spec.ts
```diff
@@ -0,0 +1,207 @@
...
+
+    beforeEach(() => {
+        spyConsoleWarn = sandbox.spy(window.console, 'warn');
+    });
+
+    afterEach(() => {
+        sandbox.restore();
+    });
+
+    describe('constructor', () => {
```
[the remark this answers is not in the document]
louis-bompart: True. 👍

src/components/ActionButton/StatefulActionButton.ts
```diff
@@ -0,0 +1,162 @@
...
+/**
+ * Represent a state that can be used by a StatefulActionButton.
+ */
+export type IStatefulActionButtonState = IActionButtonOptions & IStatefulActionButtonStateFunctionsAndUtils;
+
+interface IStatefulActionButtonStateFunctionsAndUtils {
+    /**
+     * Optional argument that could be used for logging purposes in the StatefulActionButton.
+     */
+    name: string;
```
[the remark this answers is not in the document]
louis-bompart: It is, I forgot to update the comment.

TODO: Comment not up-to-date.

tests/components/ActionButton/StatefulActionButton.spec.ts
```diff
@@ -0,0 +1,207 @@
...
+                };
+                testSubject = new StatefulActionButton(document.createElement('div'), {
+                    states: [initialState],
+                    initalState: initialState,
+                });
+            });
+
+            it('should log a warning with the name of the test', () => {
+                testSubject.switchTo(namedState);
+                expect(spyConsoleWarn.calledWithExactly('some state name is not allowed on this StatefulActionButton.'));
```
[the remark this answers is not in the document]
louis-bompart: switchTo failure warning assertions will be reworked to address the changes implied by https://github.com/coveo/search-ui-extensions/pull/84#discussion_r458754891

src/components/ActionButton/StatefulActionButton.ts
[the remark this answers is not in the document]
jeremierobert-coveo: I think the throw is a better option than being silent.

src/components/ActionButton/StatefulActionButton.ts
```diff
@@ -0,0 +1,162 @@
...
+        for (let index = 0; index < this.options.allowedTransitions.length; index++) {
+            const transition = this.options.allowedTransitions[index];
+            if (this.options.states.indexOf(transition.from) == -1) {
+                console.warn(this.generateInvalidTransitionMessage(index, true));
+                return false;
+            }
+            if (this.options.states.indexOf(transition.from) == -1) {
+                console.warn(this.generateInvalidTransitionMessage(index, false));
+                return false;
+            }
```
jeremierobert-coveo, line 133: You have twice the same condition but with different outcomes.

src/components/ActionButton/StatefulActionButton.ts
```diff
@@ -0,0 +1,162 @@
...
+        if (!this.options.initalState) {
+            console.warn('The stateful action button cannot render if the inital states is not defined.');
+            Coveo.$$(this.element).hide();
+            return false;
+        }
+        if (this.options.states.indexOf(this.options.initalState) < 0) {
+            console.warn('The stateful action button cannot render if the initial state is not in the list of states.');
+            Coveo.$$(this.element).hide();
+            return false;
+        }
```
jeremierobert-coveo, line 119: Maybe you should transform those 3 check into 3 little private function that returns a boolean

src/components/ActionButton/StatefulActionButton.ts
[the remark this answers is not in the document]
jeremierobert-coveo: Not sure about the name too ...

src/components/ActionButton/StatefulActionButton.ts
```diff
@@ -0,0 +1,162 @@
...
+
+    /**
+     * Check if a transition from the current state to @param state is allowed.
+     * @param state the destination of the transition
+     */
+    private isTransitionAllowed(state: IStatefulActionButtonState) {
+        return (
+            this.options.states.indexOf(state) > -1 &&
+            (!this.options.allowedTransitions ||
+                this.options.allowedTransitions.some((transition) => transition.from === this.currentState && transition.to === state))
```
jeremierobert-coveo, line 152: Please, create some consts for this lenghty condition.

src/components/ActionButton/StatefulActionButton.ts
```diff
@@ -0,0 +1,162 @@
...
+            if (this.options.states.indexOf(transition.from) == -1) {
+                console.warn(this.generateInvalidTransitionMessage(index, false));
+                return false;
+            }
+        }
+        return true;
+    }
+
+    private generateInvalidTransitionMessage(transitionNumber: number, isOrigin: boolean) {
+        return `The stateful action button cannot render if its origin is not in the list of states:\n\t${
```
jeremierobert-coveo, line 139: Can you store this magic template in a const or a static ?

src/components/ActionButton/StatefulActionButton.ts
```diff
@@ -0,0 +1,162 @@
...
+        return this.currentState;
+    }
+
+    /**
+     * Check if the options given to the constructor are valid.
+     * If not, it will also display the appropriate warning.
+     */
+    private areOptionsValid(): boolean {
+        if (!this.options.states?.length) {
+            console.warn('The stateful action button cannot render if no states are defined.');
```
jeremierobert-coveo, line 106: Those 3 strings should be help in a const or a static no ?

src/components/ActionButton/StatefulActionButton.ts
```diff
@@ -0,0 +1,162 @@
...
+        this.innerActionButton = new ActionButton(element, { ...this.options.initalState, click: this.handleClick.bind(this) }, bindings);
+    }
+
+    /**
+     * Switch the state of the instance if the state and the transition between the current and new state are allowed.
+     * @param state a state to try to switch to
+     */
+    public switchTo(state: IStatefulActionButtonState) {
+        if (!this.isTransitionAllowed(state)) {
+            console.warn(
```
jeremierobert-coveo, line 81: I think this magic template can be held in a const or a static

src/components/ActionButton/ToggleActionButton.ts
```diff
@@ -12,6 +12,44 @@ export interface IToggleActionButtonOptions {
...
+                this.element.setAttribute('aria-pressed', 'true');
+                if (button.options.activate) {
+                    button.options.activate.apply(button);
+                }
+            },
+            onStateExit: function () {
+                this.element.classList.remove(ToggleActionButton.ACTIVATED_CLASS_NAME);
+                this.element.setAttribute('aria-pressed', 'false');
+                if (button.options.deactivate) {
+                    button.options.deactivate.apply(this);
```
jeremierobert-coveo, line 44: You can use the look ahead operator here.
`button.options.deactivate?.apply(this);`

src/components/ActionButton/ToggleActionButton.ts
```diff
@@ -12,6 +12,44 @@ export interface IToggleActionButtonOptions {
...
+     * Create the activated state for a given ToggleActionButton
+     * @param button {ToggleActionButton}
+     */
+    static generateActivatedStateInstance(button: ToggleActionButton): IStatefulActionButtonState {
+        return {
+            onStateEntry: function () {
+                this.element.classList.add(ToggleActionButton.ACTIVATED_CLASS_NAME);
+                this.element.setAttribute('aria-pressed', 'true');
+                if (button.options.activate) {
+                    button.options.activate.apply(button);
```
jeremierobert-coveo, line 37: You can use the look ahead operator here.
`button.options.activate?.apply(this);`

src/components/ActionButton/StatefulActionButton.ts
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    /**
+     * Check if a transition from the current state to @param state is allowed.
+     * @param state the destination of the transition
+     */
+    private isTransitionAllowed(state: IStatefulActionButtonState) {
+        return (
+            this.options.states.indexOf(state) > -1 &&
+            (!this.options.allowedTransitions ||
```
jeremierobert-coveo, line 151: You can use the look ahead operator here.
`his.options.allowedTransitions?.some`

src/components/ActionButton/StatefulActionButton.ts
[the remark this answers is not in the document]
louis-bompart: It's not silent, areOptionsValid does the logging the same way that ActionButton does.
I don't mind throwing if the ActionButton throws too. But after discussing with Luc, we deem too harsh to throw altogether, a proper warning and hiding the button is a more reasonable response.

src/components/ActionButton/StatefulActionButton.ts
louis-bompart: The text of the warning is specific to this error so no, that's just cumbersome

src/components/ActionButton/StatefulActionButton.ts
louis-bompart: We don't tend to put error message in const, and I don't think there's positive value to that.

src/components/ActionButton/StatefulActionButton.ts
louis-bompart: Addressed in upcoming PR.

src/components/ActionButton/StatefulActionButton.ts
louis-bompart: No I disagree, that's already what's the method is doing by abstracting that from the constructor.
I don't like spaghetti code with a lot of functions etc. The function `areOptionsValid` does what it says and follow its comment, splitting the code would be splitting hair.

src/components/ActionButton/StatefulActionButton.ts
louis-bompart: It's in a function that does just that, no, really, **no**.

src/components/ActionButton/ToggleActionButton.ts
louis-bompart: On the second PR, you'll see that the State are actually their own full-fledged classes, which I think make this change unnecessary

src/components/ActionButton/StatefulActionButton.ts
```diff
@@ -0,0 +1,162 @@
...
+     */
+    public getCurrentState() {
+        return this.currentState;
+    }
+
+    /**
+     * Check if the options given to the constructor are valid.
+     * If not, it will also display the appropriate warning.
+     */
+    private areOptionsValid(): boolean {
```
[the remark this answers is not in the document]
louis-bompart: This function is a guard on the options. To do its job efficiently, four preconditions/guards are executed and appropriate.
> guard provides an early exit from a subroutine
https://en.wikipedia.org/wiki/Guard_(computer_science)

src/components/ActionButton/StatefulActionButton.ts
jeremierobert-coveo: It's not spaghetti code, it's well splitted functions

src/components/ActionButton/StatefulActionButton.ts
jeremierobert-coveo: I can only judge was I see

src/components/ActionButton/StatefulActionButton.ts
louis-bompart: I would agree if this was in a constructor, but I firmly believe that the net balance of splitting this private function in another three is negative.
I think the three ifs actually back-to-back showcase the nature of composite guard of the `isOptionsValid` way better than a `this.isOptionStateSet() | this.isOptionsInitialStateSet() | this.isOptionsInitialStateInOptionsStates()`.

I tried, I feel the code is less meaningful with this refactor, therefore it's not an improvement, but an hindrance and overengineering.

lbergeron: I agree with Louis that having a private method for each condition is too much split. However, when I first saw the `areOptionsValid` method, I though about asking to split the state validations in a separate private method.

We would end with `areOptionsValid` looking something like this:

    private areOptionsValid(): boolean {
        return this.areStatesValid() && this.areTransitionsValid();
    }

But honestly, I wasn't sure it was worth splitting.

tests/components/ActionButton/StatefulActionButton.spec.ts
```diff
@@ -0,0 +1,241 @@
...
+                        states: [initialState],
+                        initialState: initialState,
+                        allowedTransitions: [{ from: createSpiedState('someState')[0], to: initialState }],
+                    });
+                },
+                expectedWarningMessage:
+                    'The stateful action button cannot render if one of its transition used a state that is not in the list of states:\n\tOrigin of Transition #0 is not in the list of states. Ensure to use the same object reference as in the options.states.',
+            },
+            {
+                describe: 'if second allowedTransitions contains a transition with a origin state not included in options.state',
```
lbergeron, line 87: I am not sure about the value of performing the same test twice. Is it simply to ensure the tested code loops through all transitions?

If so, then I would suggest to keep only the test

    if second allowedTransitions contains a transition with a origin state not included in options.state

and rename it

    if allowedTransitions contains a transition with an origin state not included in options.state

tests/components/ActionButton/StatefulActionButton.spec.ts
```diff
@@ -0,0 +1,241 @@
...
+                            { from: initialState, to: initialState },
+                            { from: createSpiedState('someState')[0], to: initialState },
+                        ],
+                    });
+                },
+                expectedWarningMessage:
+                    'The stateful action button cannot render if one of its transition used a state that is not in the list of states:\n\tOrigin of Transition #1 is not in the list of states. Ensure to use the same object reference as in the options.states.',
+            },
+            {
+                describe: 'if first allowedTransitions contains a transition with a destination state not included in options.state',
```
lbergeron, line 102: Same thing as my previous comment. Unless I misunderstood the test case, I would suggest to setup two transitions and make the last one invalid. The test case should be renamed

    if allowedTransitions contains a transition with a destination state not included in options.state

src/components/ActionButton/StatefulActionButton.ts
```diff
@@ -0,0 +1,167 @@
...
+        if (!this.options.states?.length) {
+            console.warn('The stateful action button cannot render if no states are defined.');
```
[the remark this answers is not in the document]
louis-bompart: One thing I considered was the use of `console.group` and transform the guard to checks, to ensure every check is run and provide a 'full' diagnostic.
What disappointed me is that `console.group` is stuck to the info level :(

I understand where this coming from, but if we were to display the warning `'Cannot render the stateful action button because options are invalid.'` it should be done before the details (e.g. no states are defined) I think.

I'll play with this idea for a bit, but if we're to do that, the smaller checks would make sense because areOptionsValid would just play its role of boolean determination and won't do the logging anymore, meaning it would look like a 
`return condA | condB | condC |...`

src/components/ActionButton/StatefulActionButton.ts
```diff
@@ -0,0 +1,167 @@
...
+        if (this.options.states.indexOf(this.options.initialState) < 0) {
+            console.warn('The stateful action button cannot render if the initial state is not in the list of states.');
+            Coveo.$$(this.element).hide();
+            return false;
+        }
+        return !this.options.allowedTransitions || this.areTransitionsValid();
+    }
+
+    private areTransitionsValid(): boolean {
+        for (let index = 0; index < this.options.allowedTransitions.length; index++) {
```
erocheleau, line 130:
```
if (
  !this.options.allowedTransitions.every(
    (transition) =>
      this.options.states.includes(transition.from) &&
      this.options.states.includes(transition.to)
  )
) {
  console.warn(
    "All transitions of the StatefulActionButton need to be included as valid states of the button."
  );
  return false;
}
return true;
```

Ou tu veux de la compatibilite plus vieille que ca?

src/components/ActionButton/ToggleActionButton.ts
```diff
@@ -128,15 +165,7 @@ export class ToggleActionButton extends Component {
...
-            this._isActivated = activated;
-            this.updateButton();
-
-            if (this._isActivated && this.options.activate) {
-                this.options.activate();
-            }
-            if (!this._isActivated && this.options.deactivate) {
-                this.options.deactivate();
-            }
+            this.innerStatefulActionButton.switchTo(activated ? this.activatedState : this.deactivatedState);
```
erocheleau, line 168: C'est moins pire que je pensais finalement caller cette methode la.
C'est pas si pire de pas juste utiliser les 'name' des states.

src/components/ActionButton/ToggleActionButton.ts
```diff
@@ -12,6 +12,42 @@ export interface IToggleActionButtonOptions {
 }
 
 export class ToggleActionButton extends Component {
+    /**
+     * Create the deactivated state for a given ToggleActionButton
+     * @param button {ToggleActionButton}
+     */
+    static generateDeactivatedStateInstance(button: ToggleActionButton): StatefulActionButtonState {
```
erocheleau, line 19: Pourquoi le generate method versus un attribute?

src/components/ActionButton/StatefulActionButton.ts
```diff
@@ -0,0 +1,167 @@
...
+    public switchTo(state: StatefulActionButtonState) {
+        if (this.options.states.indexOf(state) === -1) {
+            console.warn(
+                `State '${state.name}' does not exist on this StatefulActionButton\nEnsure to use the object references used at the instantiation.`
+            );
+            return;
+        }
+        if (!this.isTransitionAllowed(state)) {
+            console.warn(
+                `Transition from State '${this.currentState.name}' to State '${state.name}' is not allowed on this StatefulActionButton.\nEnsure to use the object references used at the instantiation.`
```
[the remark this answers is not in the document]
louis-bompart: Attention c'est pas client facing dans le sens que le component n'est pas un JSUI component, fake si tu utilises ca sans etre un dev, you got a problem :P

src/components/ActionButton/StatefulActionButton.ts
louis-bompart: J'ai pensé, je suid d'accord qu'il y a de la complexité en plus, mais je voulais donner un log précis a l'utilisateur, car je veux pas qu'il se casse la tête a testé l'un puis l'autre facon USB A. (Is it this oriented properly, turn around, try again, fail, turn around again, succeed (maybe)

src/components/ActionButton/ToggleActionButton.ts
louis-bompart: Car au final les states seront des full-fledged class, donc pour faire un pas dans cette direction j'ai utilisé ca pour limiter le refactor dans la second PR (sneak-peak: louis-bompart/coveo-search-ui-extensions#2

tests/components/ActionButton/StatefulActionButton.spec.ts
louis-bompart: Yeah I don't remember why I pained myself to do this actually.
😕 will remove

src/components/ActionButton/ActionButton.ts
```diff
@@ -21,7 +31,7 @@ export class ActionButton extends Component {
      * The possible options for _ActionButton_.
      * @componentOptions
      */
-    static options: IActionButtonOptions = {
+    static options: ActionButtonOptions = {
```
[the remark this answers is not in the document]
louis-bompart: Ah yes this.
Ok, donc, un ActionButton doit avoir a minima soit un title, soit un icon. (having both is OK too)

Tu peux pas traduire ce genre de condition dans une interface, mais avec un union type tu peux.
D'ou ActionButtonOptions, IActionButtonOptionsWithTitle et IActionButtonOptionsWithIcon

PR #85, opened 2020-07-24
Feat/SFINT-3327: Add DisableableToggleActionButton

louis-bompart:
Add a ToggleActionButton that is disableable. (so kinda a three-states button).

Would be used by the FilterButton in SFINT


tests/components/ActionButton/DisableableToggleActionButton.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+            it('should call switchTo with the disabledState');
+        });
+
+        describe('setActivated', () => {
+            it('should do nothing if called with false');
+            it('should switch to activated if called with true');
+        });
+    });
+
+    describe('when unactivated', () => {
```
lbergeron, line 49: Is it a bad copy-paste? I guess this describe should be `when activated` and tests must be adjusted accordingly.

tests/components/ActionButton/ToggleableButton.spec.ts
```diff
@@ -0,0 +1,27 @@
+describe('ToggleUnactivatedState', () => {
```
lbergeron: You can remove the first describe block. Its test is also present in the second block.

src/components/ActionButton/DisableableToggleActionButton.ts
```diff
@@ -0,0 +1,120 @@
...
+        }
+        this.setActivated(!this.isActivated());
+
+        if (this.options.click) {
+            this.options.click();
+        }
+    }
+
+    private createInnerButton(bindings?: IResultsComponentBindings): void {
+        this.deactivatedState = new UnactivatedState(this);
```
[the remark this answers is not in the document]
louis-bompart: I thought about it, but when reviewing the class itself, it made less sense I think because the UnactivatedState is more like a 'neutral' state (neutral in the sense shift box)
I thought this reflects better the fact that there are no entry or exit functions for this state.

But both seems fine for me tho.

PR #87, opened 2020-07-28
test(SFINT-3353): Test cleaning

louis-bompart:
# What
While investigating SFINT-3353, I discovered several issues in the UT
The actual 'fix' for SFINT-3353 will be done in a second term to have a clear split between refacto' and fix.

## Improper use of `expect`:
https://jasmine.github.io/api/3.6/global.html#expect

TL;DR: Use matchers after expect. Always.

```ts
it('sometest'()=>{
   expect(false);
   expect(true);
}
```
This will pass. expect is not Assert. This will not pass:
```ts
it('sometest'()=>{
   expect(false).toBeTrue();
   expect(true).toBeTrue();
}
```

## Issue with Stub, jasmine and yield function.

Various, various issues. My goal was to, whenever possible to 'flatten' and synchronize the tests, and avoid delay as much as possible (except when used to 'flush' the promise stack see `waitForPromiseCompletions`).

## How to flatten delay

Doing an `await(delay(()=>{})` ensure that all Promises are resolved before continuing. (assuming there's no setTimeout&co going on).
This is due to how microtasks and (macro)tasks work https://developer.mozilla.org/en-US/docs/Web/API/HTML_DOM_API/Microtask_guide
> First, each time a task exits, the event loop checks to see if the task is returning control to other JavaScript code. If not, it runs all of the microtasks in the microtask queue. The microtask queue is, then, processed multiple times per iteration of the event loop, including after handling events and other callbacks.

So, long story short:
- All Promises before the `await(delay(()=>{})` are ran (if they encounter a setTimeout so be it, they add it to the (macro)task queue, which will be executed once all promises are executed. If a promise is added, however, see the article, but from what I understand, it's executed before the next (macro)task
- SetTimeout, because it's at a 0 timeout is ready for cookin' on the pile, and resolve thus resolve the Promise that returns delay, freeing up the main process and, in our case, letting the other statement executes (often the assertions).

# How did you test your tests.
By hand, most of them failed at first so, not a lot of choices (I won't say I checked _all of them_ by hand, but a good 75%)

# Further Improvements

- Apply proper unit-testing methodology by using Inversion of Control/Dependency Injections. This would allow proper 'black-boxing' and ensure that you can have more control over what's going on (e.g. with the  `ClickedDocumentList` and the `ExpendableList`)
- Isolate the display logic, the state logic, and the query logic from each other. It would make the code more readable, and allow to test -thus maintain- more easily.

### DIFF
--- a/tests/components/UserActions/UserActions.spec.ts
+++ b/tests/components/UserActions/UserActions.spec.ts
@@ -184,7 +184,7 @@ describe('UserActions', () => {
         expect(detailSection.querySelector('.CoveoUserActivity')).not.toBeNull();
     });
 
-    it('should pass the user id option to each of it sub components', async () => {
+    it('should pass the user id option to each of its sub components', async () => {
         const FAKE_USER_ID = 'someUserId' + Math.random();
 
         const automaticallyCreateComponentsInsideStub = sandbox.stub(Initialization, 'automaticallyCreateComponentsInside');
@@ -450,7 +450,7 @@ describe('UserActions', () => {
             const spyDispatchEvent = sandbox.spy(mock.cmp.element, 'dispatchEvent');
             await mock.cmp.show();
 
-            expect(spyDispatchEvent.calledOnce);
+            expect(spyDispatchEvent.calledOnce).toBeTrue();
             expect(spyDispatchEvent.firstCall.args.length).toBe(1);
             expect(spyDispatchEvent.firstCall.args[0].type).toBe('userActionsPanelShow');
         });
@@ -514,7 +514,7 @@ describe('UserActions', () => {
 
             mock.cmp.hide();
 
-            expect(spyDispatchEvent.calledOnce);
+            expect(spyDispatchEvent.calledOnce).toBeTrue();
             expect(spyDispatchEvent.firstCall.args.length).toBe(1);
             expect(spyDispatchEvent.firstCall.args[0].type).toBe('userActionsPanelHide');
         });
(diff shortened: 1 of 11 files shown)


tests/components/UserActions/UserActions.spec.ts
```diff
@@ -445,53 +410,49 @@ describe('UserActions', () => {
...
         });
 
-        it('should trigger a userActionsShow event', () => {
+        it('should trigger a userActionsShow event', async () => {
             const spyDispatchEvent = sandbox.spy(mock.cmp.element, 'dispatchEvent');
-            mock.cmp.show();
+            await mock.cmp.show();
 
-            expect(spyDispatchEvent.calledOnceWith(new CustomEvent('userActionsPanelHide')));
+            expect(spyDispatchEvent.calledOnce);
```
lbergeron, line 453: Shouldn't it be `expect(spyDispatchEvent.calledOnce).toBe(true)` instead?

tests/components/UserActions/UserActions.spec.ts
```diff
@@ -535,30 +494,29 @@ describe('UserActions', () => {
...
-        it('should trigger a userActionsHide event', () => {
-            mock.cmp.show();
+        it('should trigger a userActionsHide event', async () => {
+            await mock.cmp.show();
             const spyDispatchEvent = sandbox.spy(mock.cmp.element, 'dispatchEvent');
 
             mock.cmp.hide();
 
-            expect(spyDispatchEvent.calledOnceWith(new CustomEvent('userActionsPanelHide')));
+            expect(spyDispatchEvent.calledOnce);
```
lbergeron, line 517: Shouldn't it be `expect(spyDispatchEvent.calledOnce).toBe(true)` instead?

PR #88, opened 2020-07-28
test(SFINT-3352): Safer tests

louis-bompart:
Flaky UT issue from #86, see https://travis-ci.org/github/coveo/search-ui-extensions/builds/712557300.

I was not able to 'pin-point' with certitude what caused the flakiness, due to the issue being, well flaky (and thus I'm not 💯 confident it'll be fixed but I think it will)

Two fixes:
 - Do not modify `activateSpy` and `deactivateSpy` they're used as default options for the component instantiation, which can cause test leakage.
 - Ensure all events listener have completed their job before checking the spies. This also prevents test leakage (you could have the eventListener of the test n-1 executed during the execution of the test n) and also that we're actually testing what we want. (We do not want to 'miss' a double call because the second call occurs after the expectation.

### DIFF
--- a/tests/components/ActionButton/ToggleActionButton.spec.ts
+++ b/tests/components/ActionButton/ToggleActionButton.spec.ts
@@ -6,6 +6,7 @@ import { ActionButton } from '../../../src/components/ActionButton/ActionButton'
 import { IComponentOptions } from 'coveo-search-ui';
 import { IToggleableButtonOptions } from '../../../src/components/ActionButton/ToggleableButton';
 import { StatefulActionButton } from '../../../src/components/ActionButton/StatefulActionButton';
+import { defer } from '../../utils';
 
 describe('ToggleActionButton', () => {
     let sandbox: SinonSandbox;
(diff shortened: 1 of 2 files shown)


tests/components/ActionButton/ToggleActionButton.spec.ts
```diff
@@ -174,8 +185,11 @@ describe('ToggleActionButton', () => {
                     sandbox.reset();
                 });
 
-                it('should not call switchTo when setActivated is called with true', () => {
+                it('should not call switchTo when setActivated is called with true', async () => {
                     testSubject.setActivated(true);
+                    for (let index = 0; index < eventCompletionPromises.length; index++) {
+                        await eventCompletionPromises[index];
+                    }
```
louis-bompart, line 192: Promise.all and Jasmine cannot just cooperate (in fact, Jasmine and promises beyond basics are not working well together at all from what I experienced)

nathanlb:
Are there any metrics you noted that confirm this helps? Even just a little bit

louis-bompart:
> Are there any metrics you noted that confirm this helps? Even just a little bit

Eye-balling mostly but 10-20% failure rate before, 0% after

nathanlb:
Sounds pretty good to me

PR #90, opened 2020-08-19
fix(SFINT-3372): Fix an issue where event where incoherent

jeremierobert-coveo:
# Issue
When clicking on documents in the "Recent Clicked Documents" list, the event is sent, however it fails to log correctly in UA since there is no searchQueryUid sent. It shows up as an incoherent event with error: `CLICK_WITHOUT_ASSOCIATED_SEARCH

Since there is no search event in this scenario, we'll need to change this event to a custom event in order to properly track it.
To align with the other existing event (User Actions -> openUserActions), this would be the right values to use:
```
type: User Actions
value: userActionDocumentClick
```

# Solution
Since the issue is caused by the default behavior of the ResultLink, let's instead uses the `onClick` property of the said component to override the behavior.

This way instead logging a Click then doing the redirect, we log a custom event then we do the redirect.


tests/components/UserActions/ClickedDocumentList.spec.ts
```diff
@@ -270,6 +271,43 @@ describe('ClickedDocumentList', () => {
         expect(mock.cmp.disabled).toBe(true);
     });
 
+    it('should log a userActionDocumentClick event when a result link in the template is clicked', async () => {
```
nathanlb, line 274: this test is a bit girthy but if we aren't reusing any of this elsewhere i guess it doesn't make much sense to extract any of it either.

PR #92, opened 2020-09-15
SFINT-3372 Fix incoherent event for good

jeremierobert-coveo:
# Issue
The click event sent by `ClickedDocumentList` is not linked to any **Search Event**.

# Overview
User Actions sent by ML only have the urihash.
To display the documents, `ClickedDocumentList` fetch then by doing a search request using the `SearchEndpoint`.
That's mainly why the click event emitted by `ClickedDocumentList` are not linked to any event.

# Solutions
* Tranforming the Click event into a Custom event (We can't prevent a result link from sending a click event without modifying the Search-ui)
* Emit a Search event when we fetch the documents.

# Considerations
* We can't use the QueryController because we don't want to impact the main search.
* We don't have direct reference to the LiveAnalyticsClient.

# Flaws
The current PR manually build a search event from the QueryBuilder & the SearchRequest.

### DIFF
--- a/tests/models/UserProfilingModel.spec.ts
+++ b/tests/models/UserProfilingModel.spec.ts
@@ -166,6 +166,37 @@ describe('UserProfilingModel', () => {
             });
         });
 
+        it('should call onDocumentFetch hook when document are fetched', async () => {
+            const onDocumentFetch = sandbox.stub();
+
+            const documentResults = Fake.createFakeResults(1);
+            documentResults.results[0].raw.urihash = TEST_URI_HASH;
+
+            const endpoint = sandbox.createStubInstance(SearchEndpoint);
+            endpoint.search.callsFake(() => Promise.resolve(documentResults));
+
+            const model = new UserProfileModel(document.createElement('div'), {
+                organizationId: TEST_ORGANIZATION,
+                restUri: TEST_REST_URI,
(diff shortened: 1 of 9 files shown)


tests/models/UserProfilingModel.spec.ts
```diff
@@ -131,6 +131,25 @@ describe('UserProfilingModel', () => {
...
+                sendSearchEventsStub = sandbox.stub();
+                return {
+                    usageAnalytics: {
+                        logSearchEvent: logSearchEventStub,
+                        getPendingSearchEvent: sandbox.stub().returns({}),
+                        endpoint: {
+                            sendSearchEvents: sendSearchEventsStub,
+                        },
+                    },
+                } as any;
```
louis-bompart, line 149: Ain't that at least a Partial something?

louis-bompart (approved):
I saw the discussion in the Jira, if I understand properly, the 'product' goal is to eliminate those pesky incoherent events. If so, that seems like a good compromise.

tests/models/UserProfilingModel.spec.ts
jeremierobert-coveo: That's a test mock polyfill

tedre191 (approved):

mikegron (approved):

PR #93, opened 2020-09-17
SNOW-330 Page view links url when the id is a clickableuri

wosullivan:


src/components/UserActions/UserActivity.ts
```diff
@@ -275,9 +277,18 @@ export class UserActivity extends Component {
         const li = document.createElement('li');
         li.classList.add(VIEW_EVENT_CLASS);
 
-        const dataElement = document.createElement('div');
+        let dataElement: HTMLElement;
+        if (action.raw.content_id_key && action.raw.content_id_key === CLICKABLE_URI_ID) {
+            const a = document.createElement('a');
```
louis-bompart, line 282: Not sure about the createElement a v.s. div, that could technically break some custom CSS.
What's the motivation behind that?

src/components/UserActions/UserActions.scss
```diff
@@ -220,7 +220,8 @@ $timestamp-grey: #67768b;
             margin-left: 0.5em;
         }
 
-        .coveo-click {
+        .coveo-click,
+        .coveo-view {
```
louis-bompart, line 224: What's the reasoning behind this new rule in the context of SNOW-330?

wosullivan: It fixes the issue by wrapping the hyperlink arround with the `word-wrap: break-word` rule and keeps the same style for the different hyperlinks type.
image

src/components/UserActions/UserActivity.ts
wosullivan: I used <a> because it makes the css rules cleaner, would wrapping the <a> in a <div> be sufficient?

louis-bompart: It _should_ be alright. @jeremierobert-coveo , wdyt?

louis-bompart (approved):
LGTM overall!
Ensure @jeremierobert-coveo who worked on the feature approves it, he knows better.

src/components/UserActions/UserActivity.ts
```diff
@@ -276,8 +278,16 @@ export class UserActivity extends Component {
         li.classList.add(VIEW_EVENT_CLASS);
 
         const dataElement = document.createElement('div');
+        if (action.raw.content_id_key && action.raw.content_id_key === CLICKABLE_URI_ID) {
```
mikegron, line 281: Pas nécessaire de faire `action.raw.content_id_key &&` vue que tu valide avec `===` par la suite je pense

mikegron (approved):
Même comments que les autres, sinon looks good!

src/components/UserActions/UserActivity.ts
wosullivan: C'est vrai!

src/components/UserActions/UserActivity.ts
jeremierobert-coveo: Typically it's better to style element with classes (but customer often don't follow this rule).

src/components/UserActions/UserActivity.ts
```diff
@@ -276,8 +278,16 @@ export class UserActivity extends Component {
         li.classList.add(VIEW_EVENT_CLASS);
 
         const dataElement = document.createElement('div');
+        if (action.raw.content_id_key === CLICKABLE_URI_ID) {
```
jeremierobert-coveo, line 281: content id key is not set to permanent id by default iirc 🤔

jeremierobert-coveo (changes requested):
The more I look a the feature the more I think we should hide the Page views... (but that's an other discussion).

Also do you think that a customer should be able to change the CLICKABLE_URI_ID ?
(Maybe that won't be require, but at least we can put that in the class as a private static property. This way we can monckey patch the class in case of a maintenance case.)

src/components/UserActions/UserActivity.ts
wosullivan: Wouldn't it compare `undefined` to `clickable_uri_id` in the case that it is not present?

wosullivan:
> The more I look a the feature the more I think we should hide the Page views... (but that's an other discussion).
> 
> Also do you think that a customer should be able to change the CLICKABLE_URI_ID ?
> (Maybe that won't be require, but at least we can put that in the class as a private static property. This way we can monckey patch the class in case of a maintenance case.)

I am honestly not too familiar with the content_id_key system, but I could see the case beeing made that a customer would want to add custom content_id_keys to this?🤔

src/components/UserActions/UserActivity.ts
mikegron: @jeremierobert-coveo Je pense pas qu'ont veuille changer ca.. Ce serais pour pouvoir utiliser un autre content id key a afficher comme link ?

jeremierobert-coveo (approved):

PR #95, opened 2020-11-13
SFINT-3521: Added AugmentedResultList

nathanlb:
This component is meant to allow devs to augment their coveo results with external data before render.

### DIFF
--- a/src/components/AugmentedResultList/AugmentedResultList.ts
+++ b/src/components/AugmentedResultList/AugmentedResultList.ts
@@ -76,20 +76,20 @@ export class AugmentedResultList extends Coveo.ResultList implements IComponentB
     }
 
     private defaultMatchingFunction = (augmentData: any, queryResult: IQueryResult) => {
-        const fieldName = this.getFieldString(this.options.matchingIdField);
+        const fieldName = this.getMatchingFieldString();
         return augmentData[fieldName] === queryResult.raw[fieldName];
     };
 
     private getObjectPayload(results: IQueryResult[]): String[] {
-        const field = this.getFieldString(this.options.matchingIdField);
+        const field = this.getMatchingFieldString();
         if (results.length > 0) {
             return results.filter((result) => result.raw && result.raw[field]).map((result) => result.raw[field]);
         }
         return [];
     }
 
-    private getFieldString(fieldName: IFieldOption) {
-        return fieldName.replace('@', '');
+    private getMatchingFieldString() {
+        return this.options.matchingIdField.replace('@', '');
     }
 
     public renderResults(resultElements: HTMLElement[], append = false): Promise<void> {
@@ -99,6 +99,7 @@ export class AugmentedResultList extends Coveo.ResultList implements IComponentB
 
     public async buildResults(queryResults: Coveo.IQueryResults): Promise<HTMLElement[]> {
         let remoteResults: IPromiseReturnArgs<IAugmentData>;
+        const fieldName = this.getMatchingFieldString();
 
         if (this.options.fetchAugmentData) {
             try {
@@ -112,14 +113,14 @@ export class AugmentedResultList extends Coveo.ResultList implements IComponentB
             if (remoteResults?.data) {
                 // Merge augmenting data with Coveo Results
                 queryResults.results.forEach((res: Coveo.IQueryResult) => {
-                    const match = remoteResults.data.resultData.find((data: any) => this.options.matchingFunction(data, res));
+                    const match: any = remoteResults.data.resultData.find((data: any) => this.options.matchingFunction(data, res));
 
                     // Attach data specific to each result/object
                     for (const key in match) {
-                        if (Boolean(res.raw[key.toLowerCase()])) {
+                        if (key.toLowerCase() !== fieldName && Boolean(res.raw[key.toLowerCase()])) {
                             this.logger.warn(`The ${key} field was overwritten on result: ${res.title}`);
                         }
-                        res.raw[key.toLowerCase()] = (match as any)[key];
+                        res.raw[key.toLowerCase()] = match[key];
                     }
 
                     // Attach data common to all results
--- a/src/components/AugmentedResultList/AugmentedResultList.ts
+++ b/src/components/AugmentedResultList/AugmentedResultList.ts
@@ -1,36 +1,36 @@
-import { $$, ComponentOptions, IComponentBindings, Initialization, IQueryResult, IResultListOptions, IFieldOption } from 'coveo-search-ui';
+import { ComponentOptions, IComponentBindings, Initialization, IQueryResult, IResultListOptions, IFieldOption } from 'coveo-search-ui';
 
 /**
- * Interface for the object data returned from remote action.
+ * Interface for the data returned from external fetch action.
  *
  * @export
- * @interface IObjectData
+ * @interface IAugmentData
  */
-export interface IObjectData {
+export interface IAugmentData {
     /**
      * Data specific to a result with matching object id.
      */
-    objectData: {}[];
+    resultData: any[];
     /**
      * Data to add to every result with matching object id.
      */
-    commonData: {};
+    commonData: any;
 }
 
 /**
- * Generic interface for the response returned by the remote action method.
+ * Generic interface for the response returned by the external fetch action.
  *
  * @export
  * @interface IPromiseReturnArgs
  * @template T
  */
-export interface IPromiseReturnArgs {
-    data: IObjectData;
+export interface IPromiseReturnArgs<T> {
+    data: T;
 }
 
 export interface AugmentedResultListOptions extends IResultListOptions {
     matchingIdField: IFieldOption;
-    objectDataAction: (objectIds: String[]) => Promise<IPromiseReturnArgs>;
+    fetchAugmentData: (objectIds: String[]) => Promise<IPromiseReturnArgs<IAugmentData>>;
 }
 
 export class AugmentedResultList extends Coveo.ResultList implements IComponentBindings {
@@ -41,15 +41,15 @@ export class AugmentedResultList extends Coveo.ResultList implements IComponentB
      */
     static options: AugmentedResultListOptions = {
         /**
-         * The field to be used as matching ID between object and result.
+         * The field to be used as matching ID between augment data and result.
          */
         matchingIdField: ComponentOptions.buildFieldOption({
             required: true,
         }),
         /**
-         * The function used to fetch extra object information.
+         * The function used to fetch extra result information.
          */
-        objectDataAction: ComponentOptions.buildCustomOption<(objectIds: String[]) => Promise<IPromiseReturnArgs>>(() => {
+        fetchAugmentData: ComponentOptions.buildCustomOption<(objectIds: String[]) => Promise<IPromiseReturnArgs<IAugmentData>>>(() => {
             return null;
         }),
     };
@@ -77,49 +77,36 @@ export class AugmentedResultList extends Coveo.ResultList implements IComponentB
         return fieldName.replace('@', '');
     }
 
-    protected enableAnimation() {
-        if (!document.getElementById('overlay')) {
-            $$(document.body).append($$('div', { id: 'overlay', class: 'modal-backdrop fade in' }).el);
-        }
-    }
-
-    protected disableAnimation() {
-        if (document.getElementById('overlay')) {
-            document.getElementById('overlay').remove();
-        }
-    }
-
     public renderResults(resultElements: HTMLElement[], append = false): Promise<void> {
         const res = super.renderResults(resultElements, append);
-        this.disableAnimation();
         return res;
     }
 
-    public async buildResults(results: Coveo.IQueryResults): Promise<HTMLElement[]> {
-        this.enableAnimation();
+    public async buildResults(queryResults: Coveo.IQueryResults): Promise<HTMLElement[]> {
         const fieldString = this.getFieldString(this.options.matchingIdField);
 
-        if (this.options.objectDataAction) {
-            // Call remote action to fetch object data
-            const remoteResults: IPromiseReturnArgs = await this.options
-                .objectDataAction(this.getObjectPayload(results.results))
-                .then((data) => {
-                    return data;
-                })
-                .catch((ex) => {
-                    this.logger.error('Unable to fetch object data.');
-                    return null;
-                });
+        if (this.options.fetchAugmentData) {
+            let remoteResults: IPromiseReturnArgs<IAugmentData>;
+            try {
+                // Call remote action to fetch augmenting data
+                remoteResults = await this.options.fetchAugmentData(this.getObjectPayload(queryResults.results));
+            } catch (e) {
+                this.logger.error(['Unable to fetch augment data.', e]);
+                return null;
+            }
 
-            if (remoteResults && remoteResults.data) {
+            if (remoteResults?.data) {
                 // Merge remote action results with Coveo Results
-                results.results.forEach((res: Coveo.IQueryResult) => {
-                    const match = remoteResults.data.objectData.find((data) => {
+                queryResults.results.forEach((res: Coveo.IQueryResult) => {
+                    const match = remoteResults.data.resultData.find((data) => {
                         return (data as any)[fieldString] === res.raw[fieldString];
                     });
 
                     // Attach data specific to each result/object
                     for (const key in match) {
+                        if (Boolean(res.raw[key.toLowerCase()])) {
+                            this.logger.warn(`The ${key} field was overwritten on result: ${res.title}`);
+                        }
                         res.raw[key.toLowerCase()] = (match as any)[key];
                     }
 
@@ -133,7 +120,7 @@ export class AugmentedResultList extends Coveo.ResultList implements IComponentB
             this.logger.error('No objectDataAction is defined.');
         }
 
-        const ret = super.buildResults(results);
+        const ret = super.buildResults(queryResults);
         return ret;
     }
 }
--- a/src/components/AugmentedResultList/AugmentedResultList.ts
+++ b/src/components/AugmentedResultList/AugmentedResultList.ts
@@ -48,9 +48,12 @@ export class AugmentedResultList extends Coveo.ResultList implements IComponentB
         /**
          * The function used to fetch extra result information.
          */
-        fetchAugmentData: ComponentOptions.buildCustomOption<(objectIds: String[]) => Promise<IPromiseReturnArgs<IAugmentData>>>(() => {
-            return null;
-        }, {required: true}),
+        fetchAugmentData: ComponentOptions.buildCustomOption<(objectIds: String[]) => Promise<IPromiseReturnArgs<IAugmentData>>>(
+            () => {
+                return null;
+            },
+            { required: true }
+        ),
         /**
          * The function to use to determine a match between augment data and query results.
          */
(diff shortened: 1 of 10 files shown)


src/components/AugmentedResultList/AugmentedResultList.ts
```diff
@@ -0,0 +1,142 @@
+import { $$, ComponentOptions, IComponentBindings, Initialization, IQueryResult, IResultListOptions, IFieldOption } from 'coveo-search-ui';
+
+/**
+ * Interface for the object data returned from remote action.
```
louis-bompart, line 4: I don't think we should refer to remote action, that's a Salesforce specific terminology.

src/components/AugmentedResultList/AugmentedResultList.ts
```diff
@@ -0,0 +1,142 @@
...
+/**
+ * Generic interface for the response returned by the remote action method.
+ *
+ * @export
+ * @interface IPromiseReturnArgs
+ * @template T
+ */
+export interface IPromiseReturnArgs {
+    data: IObjectData;
+}
```
louis-bompart, line 29: I think the name of those interfaces is a bit unprecise, I would maybe prefix them with 'ResultListAugmenter' (augmenter as https://en.wiktionary.org/wiki/augmenter not, the french word for 'increase').

I'm not sure about ResultListAugmenter but, something of the like prolly.

src/components/AugmentedResultList/AugmentedResultList.ts
```diff
@@ -0,0 +1,142 @@
...
+     */
+    commonData: {};
+}
+
+/**
+ * Generic interface for the response returned by the remote action method.
+ *
+ * @export
+ * @interface IPromiseReturnArgs
+ * @template T
```
louis-bompart, line 25: Where's the template? (that's actually smart to use a template btw, you could templatize IObjectData too so it less loose than '{}')

src/components/AugmentedResultList/AugmentedResultList.ts
```diff
@@ -0,0 +1,142 @@
...
+        return [];
+    }
+
+    private getFieldString(fieldName: IFieldOption) {
+        return fieldName.replace('@', '');
+    }
+
+    protected enableAnimation() {
+        if (!document.getElementById('overlay')) {
+            $$(document.body).append($$('div', { id: 'overlay', class: 'modal-backdrop fade in' }).el);
```
louis-bompart, line 82: This is really odd, and I have doubt that it'll work well in Salesforce because we cannot 'escape' our components.

Could you explain what this does, so we can find an alternative?
(Also, I would avoid using `$$` when you can use native methods ;) )

src/components/AugmentedResultList/AugmentedResultList.ts
```diff
@@ -0,0 +1,142 @@
...
+        }
+    }
+
+    protected disableAnimation() {
+        if (document.getElementById('overlay')) {
+            document.getElementById('overlay').remove();
+        }
```
louis-bompart, line 89: How about a oneliner?
```suggestion
        document.getElementById('overlay')?.remove();
```
(Yep, it works after a function call was not sure either, see [TS Playground here](https://www.typescriptlang.org/play?#code/MYewdgzgLgBAtgTwGIFczCgS3DAvDACgCcBTKFIsAZTigAsBzALhgCMQQAbEgQzAEo8APhilylGvQYwA-DADeMAA5EQSlgX64hAchB0EPHTAC+MFmgAmJAGaYwJSwG4AUKEhcSAOk4gGBRFR0LHACGx5OCBJ+GS8VNU1+IA))

src/components/AugmentedResultList/AugmentedResultList.ts
```diff
@@ -0,0 +1,142 @@
...
+                .objectDataAction(this.getObjectPayload(results.results))
+                .then((data) => {
+                    return data;
+                })
+                .catch((ex) => {
+                    this.logger.error('Unable to fetch object data.');
+                    return null;
+                });
+
+            if (remoteResults && remoteResults.data) {
```
louis-bompart, line 114: Could be simplified with optional chaining.
```suggestion
            if (remoteResults?.data) {
```

src/components/AugmentedResultList/AugmentedResultList.ts
```diff
@@ -0,0 +1,142 @@
...
+        }
+    }
+
+    public renderResults(resultElements: HTMLElement[], append = false): Promise<void> {
+        const res = super.renderResults(resultElements, append);
+        this.disableAnimation();
+        return res;
+    }
+
+    public async buildResults(results: Coveo.IQueryResults): Promise<HTMLElement[]> {
```
louis-bompart, line 98: I would rename results by queryResults to avoid `results.results` which is odd to read.

louis-bompart (changes requested):
Some minor improvements and some doubts about the `document.body`.

src/components/AugmentedResultList/AugmentedResultList.ts
```diff
@@ -0,0 +1,142 @@
...
+        /**
+         * The field to be used as matching ID between object and result.
+         */
+        matchingIdField: ComponentOptions.buildFieldOption({
+            required: true,
+        }),
+        /**
+         * The function used to fetch extra object information.
+         */
+        objectDataAction: ComponentOptions.buildCustomOption<(objectIds: String[]) => Promise<IPromiseReturnArgs>>(() => {
```
jeremierobert-coveo, line 52: Not sure about the naming here. Do you think there is a name that can describe the action without looking at the comment ?

src/components/AugmentedResultList/AugmentedResultList.ts
jeremierobert-coveo: Iirc that's from the original result list

src/components/AugmentedResultList/AugmentedResultList.ts
```diff
@@ -0,0 +1,142 @@
...
+        this.enableAnimation();
+        const fieldString = this.getFieldString(this.options.matchingIdField);
+
+        if (this.options.objectDataAction) {
+            // Call remote action to fetch object data
+            const remoteResults: IPromiseReturnArgs = await this.options
+                .objectDataAction(this.getObjectPayload(results.results))
+                .then((data) => {
+                    return data;
+                })
```
jeremierobert-coveo, line 108: you can replace this by
```suggestion
                .then((data) => data)
```

src/components/AugmentedResultList/AugmentedResultList.ts
nathanlb: Haha now i'm embarrassed. I took this from the class PS made without thinking of changing it citing: "if it ain't broke, don't fix it". Should've fixed it :P

src/components/AugmentedResultList/AugmentedResultList.ts
```diff
@@ -0,0 +1,142 @@
...
+        return res;
+    }
+
+    public async buildResults(results: Coveo.IQueryResults): Promise<HTMLElement[]> {
+        this.enableAnimation();
+        const fieldString = this.getFieldString(this.options.matchingIdField);
+
+        if (this.options.objectDataAction) {
+            // Call remote action to fetch object data
+            const remoteResults: IPromiseReturnArgs = await this.options
```
jeremierobert-coveo, line 104: Why do you then catch while you are using await 😕 ?

src/components/AugmentedResultList/AugmentedResultList.ts
```diff
@@ -0,0 +1,142 @@
...
+
+            if (remoteResults && remoteResults.data) {
+                // Merge remote action results with Coveo Results
+                results.results.forEach((res: Coveo.IQueryResult) => {
+                    const match = remoteResults.data.objectData.find((data) => {
+                        return (data as any)[fieldString] === res.raw[fieldString];
+                    });
+
+                    // Attach data specific to each result/object
+                    for (const key in match) {
```
jeremierobert-coveo, line 122: Can you change the for loops for their functionnal variant ? (Object.keys().forEach)

jeremierobert-coveo (changes requested):

src/components/AugmentedResultList/AugmentedResultList.ts
```diff
@@ -0,0 +1,142 @@
...
+    static ID = 'AugmentedResultList';
+
+    /**
+     * @componentOptions
+     */
+    static options: AugmentedResultListOptions = {
+        /**
+         * The field to be used as matching ID between object and result.
+         */
+        matchingIdField: ComponentOptions.buildFieldOption({
```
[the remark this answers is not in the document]
nathanlb: In my local branch here i have a warning message that logs when the field value is replaced. This should make the cause fairly easy to spot if a problem arises. To have a generic component like this I think it needs to be fairly loose. So the responsibility would be on the documentation and the user of the component to exercise caution. Do you have another idea?

src/components/AugmentedResultList/AugmentedResultList.ts
nathanlb: I wouldn't mind doing it but what's the problem you see with this?

src/components/AugmentedResultList/AugmentedResultList.ts
nathanlb: I assumed PS had their reasons. I can change it

src/components/AugmentedResultList/AugmentedResultList.ts
nathanlb: I gets a bit messier tbh...
`remoteResults.data.commonData.keys().forEach((key: any) => {`
instead of
`for (const key in remoteResults.data.commonData) {`

src/components/AugmentedResultList/AugmentedResultList.ts
louis-bompart: If you can try to 'flatten' the code by transforming the then catch by a try catch, it'd be better imho.

nathanlb: Yep that's what i went for

src/components/AugmentedResultList/AugmentedResultList.ts
louis-bompart: I checked the code of ResultList.ts, I don't think so. @nathanlb, I'd double-check this while reworking on the PR if I were you ;)

src/components/AugmentedResultList/AugmentedResultList.ts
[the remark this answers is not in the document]
jeremierobert-coveo: Having a comparision method as a parameter instead ? This will allow more complex cases and it's still simple enough to easily cover the base schenario.

src/components/AugmentedResultList/AugmentedResultList.ts
jeremierobert-coveo: the for in loop had some issue in IE 11 iirc.

louis-bompart: @jeremierobert-coveo you're probably mixing up the for..of and the for..in:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in

nathanlb: correct. it's for...of that causes issues.

nathanlb: I want to leave it as is

src/components/AugmentedResultList/AugmentedResultList.ts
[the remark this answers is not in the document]
nathanlb: Not a bad idea. i'll test it out

src/components/AugmentedResultList/AugmentedResultList.ts
nathanlb: I thought it was too. but it isn't so i removed it. Must have to do with the ecolab custo. The display behaviour is fine without it

src/components/AugmentedResultList/AugmentedResultList.ts
```diff
@@ -0,0 +1,141 @@
...
+     * @param element The HTMLElement on which to instantiate the component.
+     * @param options The options for the `ResultList` component.
+     * @param bindings The bindings that the component requires to function normally. If not set, these will be
+     * automatically resolved (with a slower execution time).
+     */
+    constructor(public element: HTMLElement, public options: AugmentedResultListOptions, public bindings: IComponentBindings) {
+        super(element, ComponentOptions.initComponentOptions(element, AugmentedResultList, options), bindings, AugmentedResultList.ID);
+        if (!this.options.matchingFunction) {
+            this.options.matchingFunction = this.defaultMatchingFunction;
+        }
```
jeremierobert-coveo, line 75: You can replace this with a one liner
```suggestion
        this.options.matchingFunction = this.options.matchingFunction ?? this.defaultMatchingFunction;
```

src/components/AugmentedResultList/AugmentedResultList.ts
```diff
@@ -0,0 +1,141 @@
...
+        const fieldName = this.getFieldString(this.options.matchingIdField);
+        return augmentData[fieldName] === queryResult.raw[fieldName];
+    };
+
+    private getObjectPayload(results: IQueryResult[]): String[] {
+        const field = this.getFieldString(this.options.matchingIdField);
+        if (results.length > 0) {
+            return results.filter((result) => result.raw && result.raw[field]).map((result) => result.raw[field]);
+        }
+        return [];
```
jeremierobert-coveo, line 88: The if is redundant. You can simplify using
```suggestion
        return results.filter((result) => result.raw && result.raw[field]).map((result) => result.raw[field]);
```

src/components/AugmentedResultList/AugmentedResultList.ts
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    private getFieldString(fieldName: IFieldOption) {
+        return fieldName.replace('@', '');
+    }
+
+    public renderResults(resultElements: HTMLElement[], append = false): Promise<void> {
+        const res = super.renderResults(resultElements, append);
+        return res;
```
jeremierobert-coveo, line 97: the const is not required.
```suggestion
        return super.renderResults(resultElements, append);
```

src/components/AugmentedResultList/AugmentedResultList.ts
```diff
@@ -0,0 +1,141 @@
...
+                        res.raw[key.toLowerCase()] = (remoteResults.data.commonData as any)[key];
+                    }
+                });
+            }
+        } else {
+            this.logger.error('No objectDataAction is defined.');
+        }
+
+        const ret = super.buildResults(queryResults);
+        return ret;
```
jeremierobert-coveo, line 136:
```suggestion
        return super.buildResults(queryResults);
```

jeremierobert-coveo (changes requested):
I think doc should pass over the component.

src/components/AugmentedResultList/AugmentedResultList.ts
```diff
@@ -0,0 +1,141 @@
...
+export interface IPromiseReturnArgs<T> {
+    data: T;
+}
+
+export interface AugmentedResultListOptions extends IResultListOptions {
+    matchingIdField: IFieldOption;
+    fetchAugmentData: (objectIds: String[]) => Promise<IPromiseReturnArgs<IAugmentData>>;
+    matchingFunction: (augmentData: any, queryResult: IQueryResult) => boolean;
+}
+
```
jpmarceau, line 36: Could be nice to add a doc string to link to `[Using the Augmented Result List Component](http://docs.coveo.com/en/3421)`, which will be published when [this PR](https://bitbucket.org/coveord/doc_jekyll-public-site/pull-requests/4748/sfint-3523-augmented-result-list/diff) is merged.

nathanlb: I'm not so sure about that one since this component is meant to be agnostic to any specific platform like salesforce.

jpmarceau: fair enough 🙃

src/components/AugmentedResultList/AugmentedResultList.ts
```diff
@@ -0,0 +1,135 @@
...
+            if (remoteResults?.data) {
+                // Merge augmenting data with Coveo Results
+                queryResults.results.forEach((res: Coveo.IQueryResult) => {
+                    const match: any = remoteResults.data.resultData.find((data: any) => this.options.matchingFunction(data, res));
+
+                    // Attach data specific to each result/object
+                    for (const key in match) {
+                        if (key.toLowerCase() !== fieldName && Boolean(res.raw[key.toLowerCase()])) {
+                            this.logger.warn(`The ${key} field was overwritten on result: ${res.title}`);
```
lbergeron, line 115: I guess we should log a debug trace instead of a warning as it is part of the expected component behavior.

jeremierobert-coveo (approved):

src/components/AugmentedResultList/AugmentedResultList.ts
```diff
@@ -0,0 +1,135 @@
...
+                        }
+                        res.raw[key.toLowerCase()] = match[key];
+                    }
+
+                    // Attach data common to all results
+                    for (const key in remoteResults.data.commonData) {
```
lbergeron, line 121: I think we should merge the common data before the specific one. If we happen to have the same field in both common and specific data, I would expect the specific value to supersede the common one.

lbergeron (commented):
Please validate my concern about precedence when merging augmented data.

nathanlb:
> Please validate my concern about precedence when merging augmented data.

I think you're correct that the specific data should take precedence.

tests/components/AugmentedResultList/AugmentedResultList.spec.ts
```diff
@@ -0,0 +1,219 @@
...
+            const displayedResults = test.cmp.getDisplayedResults();
+            expect(displayedResults.length).toEqual(numResults);
+            expect(displayedResults.find((result) => result.raw.id === '#007').raw.name).toEqual('squirtle');
+            expect(displayedResults.find((result) => result.raw.id === '#001').raw.name).toBeUndefined();
+            expect(displayedResults.find((result) => result.raw.id === '#004').raw.name).toBeUndefined();
+            done();
+        });
+    });
+
+    it('should augment results with object data and warn if an overwrite occured', (done) => {
```
lbergeron, line 166: Adjust the test name with the correct logging level

tests/components/AugmentedResultList/AugmentedResultList.spec.ts
```diff
@@ -0,0 +1,219 @@
...
+        const test = createComponent(stubFetchAugmentData);
+
+        test.cmp.buildResults(data).then(() => {
+            expect(test.cmp.getDisplayedResults().length).toEqual(numResults);
+            verifyAugmentedResults(returnData, test.cmp.getDisplayedResults());
+            done();
+        });
+    });
+
+    it('should augment results with object data using provided matching function', (done) => {
```
lbergeron, line 151: It would be nice to have a test to validate that precedence between common and specific data is applied properly.

lbergeron (approved):

tests/components/AugmentedResultList/AugmentedResultList.spec.ts
nathanlb: done 👍

louis-bompart (approved):
LGTM, RIP Lazy, except if @mikegron sees an inconvenience with that.

PR #100, opened 2022-01-25
build(jenkins): add air fresheners on build

louis-bompart:
Does it work? Yeah.
Is it good? Nah, but the time required to make it good is way more than I can spend on that.

- Change from `node-sass` to `sass`: node-sass use native dep, and it was broken. Moreover, node-sass is deprecated in favour of `sass` and is a drop-in replacement (yes, the drop-in API are deprecated, but I will not venture any further than the make it work stage')

- Snyk: I don't see the point of taking the DepPipeline for that, and I had difficulties polling the ECR to get it, thus the change

- New env variable + few commands to circumvent issue while running as `root`, which was simpler than not running as root.


erocheleau (approved):
Hey that looks perfect to me!
I was expecting more hacks, seems perfect to define the folders first

mikegron (approved):
Crime, c'est plus clean que j'aurais penser haha!

louis-bompart:
I think we stopped using Travis 🤔 @mikegron or @erocheleau , any one of you can remove the check in the repo setting?

PR #103, opened 2022-02-01
Feat/sfint 4378 added a tooltip to show origin1 on clickedDocList and queriesList

erocheleau:
image
image

### DIFF
--- a/src/components/UserActions/ClickedDocumentList.ts
+++ b/src/components/UserActions/ClickedDocumentList.ts
@@ -1,15 +1,4 @@
-import {
-    Component,
-    IComponentBindings,
-    Initialization,
-    ComponentOptions,
-    IQueryResult,
-    Template,
-    HtmlTemplate,
-    QueryUtils,
-    l,
-    get,
-} from 'coveo-search-ui';
+import { Component, IComponentBindings, Initialization, ComponentOptions, Template, HtmlTemplate, QueryUtils, l, get } from 'coveo-search-ui';
 import { UserProfileModel, UserAction } from '../../models/UserProfileModel';
 import { ExpandableList } from './ExpandableList';
 import { UserActionType } from '../../rest/UserProfilingEndpoint';
@@ -19,7 +8,7 @@ import './Strings';
 /**
  * Initialization options of the **ClickedDocumentList** class.
  */
-export interface IClickedDocumentList {
+export interface IClickedDocumentListOptions {
     /**
      * Number of Clicked Documents shown.
      *
@@ -61,7 +50,7 @@ export class ClickedDocumentList extends Component {
     /**
      * Default initialization options of the **ClickedDocumentList** class.
      */
-    static readonly options: IClickedDocumentList = {
+    static readonly options: IClickedDocumentListOptions = {
         numberOfItems: ComponentOptions.buildNumberOption({
             defaultValue: 3,
             min: 1,
@@ -84,7 +73,7 @@ export class ClickedDocumentList extends Component {
     };
 
     private userProfileModel: UserProfileModel;
-    private sortedDocumentsList: IQueryResult[];
+    private sortedDocumentsList: UserAction[];
 
     /**
      * Create an instance of **ClickedDocumentList**. Initialize is needed the **UserProfileModel** and fetch user actions related to the **UserId**.
@@ -93,7 +82,7 @@ export class ClickedDocumentList extends Component {
      * @param options Initialization options of the component.
      * @param bindings Bindings of the Search-UI environment.
      */
-    constructor(public element: HTMLElement, public options: IClickedDocumentList, public bindings: IComponentBindings) {
+    constructor(public element: HTMLElement, public options: IClickedDocumentListOptions, public bindings: IComponentBindings) {
         super(element, ClickedDocumentList.ID, bindings);
 
         this.options = ComponentOptions.initComponentOptions(element, ClickedDocumentList, options);
@@ -113,7 +102,7 @@ export class ClickedDocumentList extends Component {
                 .reduce(this.filterDuplicatesClickAction, [])
                 .map((action) => {
                     action.document.searchInterface = this.searchInterface;
-                    return action.document;
+                    return action;
                 });
             this.render();
         }, this.logger.error.bind(this.logger));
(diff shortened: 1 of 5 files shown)


src/components/UserActions/QueryList.ts
```diff
@@ -156,6 +161,18 @@ export class QueryList extends Component {
         }
         return listItem;
     }
+
+    private addTooltipElement(element: HTMLElement, action: UserAction) {
```
mmitiche, line 165: would it make sense to isolate the logic of this method in a separate file avoid writing  it twice?

src/components/UserActions/UserActions.scss
```diff
@@ -116,6 +117,30 @@ $clickable-blue: #004990;
...
+                    border-radius: 4px;
+                    color: white;
+                    margin-left: 1.5em;
+
+                    &:after {
+                        position: absolute;
+                        top: 100%;
+                        content: '';
+                        left: 1em;
+                        border: 10px solid $clickable-blue;
```
mmitiche, line 136: We can get rid of $clickable-blue here no? to have just
 `border: 10px solid;`  and the color will keep its default value here before being overwritten in the line after.

mmitiche: Or maybe having soemthing like this:
```
  border: 10px solid transparent;
  border-top-color: #004990;
```

mmitiche (approved):
Very Nice, the position of the triangle is so natural 👌

src/components/UserActions/ClickedDocumentList.ts
```diff
@@ -124,19 +113,22 @@ export class ClickedDocumentList extends Component {
...
+                QueryUtils.setSearchInterfaceObjectOnQueryResult(this.searchInterface, action.document);
+                return (<Promise<HTMLElement>>this.options.template.instantiateToElement(action.document, {
                     wrapInDiv: true,
                     checkCondition: true,
                     currentLayout: 'list',
                     responsiveComponents: this.searchInterface.responsiveComponents,
                 })).then((element) => {
-                    Initialization.automaticallyCreateComponentsInsideResult(element, result);
+                    Initialization.automaticallyCreateComponentsInsideResult(element, action.document);
+                    if (action.raw.origin_level_1) {
```
lbergeron, line 129: The issue description says:

> Add a tooltip on each item of the session summary to display the origin_level_1 or the search-hub of the items

But, we are only checking for `origin_level_1` here. Is it because `origin_level_1` defaults on the search hub?

src/components/UserActions/ClickedDocumentList.ts
```diff
@@ -146,6 +138,18 @@ export class ClickedDocumentList extends Component {
             showLessMessage: l(`${ClickedDocumentList.ID}_less`),
         });
     }
+
+    private addTooltipElement(element: HTMLElement, action: UserAction) {
+        const insertBeforeElement = element.querySelector('.CoveoResultLink');
+        if (insertBeforeElement) {
+            const tooltip = document.createElement('div');
+            tooltip.classList.add('coveo-tooltip-origin1');
+            tooltip.innerText = action.raw.origin_level_1;
```
lbergeron, line 147: Are you sure it would be clear to agents that the tooltip content is the origin level 1? I think it depends a lot on whether the origin level 1 value is meaningful.

What do you think?

src/components/UserActions/ClickedDocumentList.ts
```diff
@@ -19,7 +8,7 @@ import './Strings';
 /**
  * Initialization options of the **ClickedDocumentList** class.
  */
-export interface IClickedDocumentList {
+export interface IClickedDocumentListOptions {
```
lbergeron, line 11: Renaming this interface is a breaking change. I understand the new name is better and this library has not yet reached version 1.0.

However, it might still cause issues to developers when upgrading.

I'd suggest to either:

1. revert the change to make it non-breaking
2. or, make sure to document the breaking change in the merge commit description. Hopefully, it would help a developer facing the issue to fix the code.

src/components/UserActions/QueryList.ts
erocheleau: Yea I thought about it, felt a bit weird to pass in an HTMLElement belonging to this component to another file, and just extracting the logic purely for the tooltip made for very little code to put in another file

src/components/UserActions/ClickedDocumentList.ts
erocheleau: origin level 1 is the same thing as the searchhub yes, I put searchhub here because the terms are interchangeable most of the time people don't know that searchhub is actually originLevel1 as far as analytics is concerned

src/components/UserActions/ClickedDocumentList.ts
erocheleau: I followed the mockup for that, but I think the original logic is that agents don't know more what a "search hub" or a "origin" is really, so this is more or less debug info for the educated and these people would know what this value is.

I could be mis-remembering the reasoning though.

src/components/UserActions/ClickedDocumentList.ts
erocheleau: oh yea very true, I'll just revert that change 👍 
I don't put enough value in changing the name I think to deserve creating a breaking change.

src/components/UserActions/QueryList.ts
erocheleau: Also the selector for the insertBeforeElement
`const insertBeforeElement = element.querySelector('.coveo-link');`

changes between the two files.

src/components/UserActions/UserActions.scss
erocheleau: yup you are right

src/components/UserActions/ClickedDocumentList.ts
lbergeron: > ...so this is more or less debug info for the educated and these people would know what this value is

I'm not sure using a tooltip to display this information is the best choice then. My understanding is that the tooltip will be confusing for a majority of agents. To be honest, I think it will add noise for most of them.

@adroletCoveo Instead of using a tooltip, what would you think of adding an "information" icon or something that the agent could click to reveal a popup with the details? It would give some more space to indicate what this value is. More troubleshooting values could be added in the future.

Another possibility could be to have some kind of switch to enable this kind of troubleshooting information, so agents are not confused with it, and others can still access it quickly.

Any thoughts?

erocheleau: No problem with these suggestions, but I would advocate they are definitely out of scope for this feature that was already planned and designed.

Perhaps in the Quantic version?

adroletCoveo: Very good point @lbergeron  I agree that the tooltips adds noise. You have good suggestions, my only concern with the information icon is that repeating it might also add noise. The switch idea could be nice, since it's useful to only a few. I'll investigate it for the Quantic version.  
 
For this version I'll check with Antoine since he's the one who worked on the Lightning improvements, maybe the trade-off noise/usefulness is not worth it.

adroletCoveo: Just to mention for documenting purposes, after speaking with Étienne this morning, Antoine and I think it would be better not to show the tooltip. We believe so because so far we haven't found the reason why it was needed in the first place. However it is not a terrible experience for the user and could be left as is. (but we prefer not to ;) ) 

We agreed with Étienne to follow what PMs prefer depending on the need vs time required. 
But I will think about this issue for the Quantic version of the Insight panel.

PR #106, opened 2022-02-03
feat(SFINT-4388) no results text when no user actions

erocheleau:
New way of explaining why there are no user actions visible.

Where the user actions are not enabled (404 response) :
image

When we get an error or no user actions:
image

### DIFF
--- a/src/components/UserActions/Strings.ts
+++ b/src/components/UserActions/Strings.ts
@@ -1,7 +1,13 @@
 import { Translation, Language } from '../../utils/translation';
 
 Translation.register(Language.English, {
-    UserActions_no_actions: 'No actions available for this user',
+    UserActions: 'User Actions',
+    UserActions_no_actions_title: 'No actions available for this user',
+    UserActions_no_actions_causes_title: 'Potential causes',
+    UserActions_no_actions_cause_not_enabled: 'User actions are not enabled',
+    UserActions_no_actions_cause_not_associated: 'There are no user actions associated with the user',
+    UserActions_no_actions_cause_case_too_old: 'The case is too old to detect related actions',
+    UserActions_no_actions_contact_admin: 'Contact your administrator for help',
     UserActions_enable_prompt: 'The User Action feature is not activated for your organization.\nTo activate it, contact Coveo Support.',
 
     QueryList_more: 'Show More',
--- a/src/components/UserActions/Strings.ts
+++ b/src/components/UserActions/Strings.ts
@@ -4,7 +4,7 @@ Translation.register(Language.English, {
     UserActions: 'User Actions',
     UserActions_no_actions_title: 'No actions available for this user',
     UserActions_no_actions_causes_title: 'Potential causes',
-    UserActions_no_actions_cause_not_enabled: 'User actions are not enabled',
+    UserActions_no_actions_cause_not_enabled: 'User actions are not enabled for your organization',
     UserActions_no_actions_cause_not_associated: 'There are no user actions associated with the user',
     UserActions_no_actions_cause_case_too_old: 'The case is too old to detect related actions',
     UserActions_no_actions_contact_admin: 'Contact your administrator for help',
--- a/src/components/UserActions/UserActions.ts
+++ b/src/components/UserActions/UserActions.ts
@@ -315,9 +315,9 @@ export class UserActions extends Component {
     }
 
     private renderNoActions() {
-        const element = document.createElement('div');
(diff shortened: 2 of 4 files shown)


src/components/UserActions/Strings.ts
```diff
@@ -1,7 +1,13 @@
...
 
 Translation.register(Language.English, {
-    UserActions_no_actions: 'No actions available for this user',
+    UserActions: 'User Actions',
+    UserActions_no_actions_title: 'No actions available for this user',
+    UserActions_no_actions_causes_title: 'Potential causes',
+    UserActions_no_actions_cause_not_enabled: 'User actions are not enabled for your organization',
+    UserActions_no_actions_cause_not_associated: 'There are no user actions associated with the user',
+    UserActions_no_actions_cause_case_too_old: 'The case is too old to detect related actions',
+    UserActions_no_actions_contact_admin: 'Contact your administrator for help',
```
erocheleau, line 10: @acarpentier01 These are the strings. you can check on the screenshots where they are used.

acarpentier01 (approved):
Looks great! 👍

adroletCoveo (approved):
All good for me! Thanks!

src/components/UserActions/UserActions.ts
```diff
@@ -318,16 +317,33 @@ export class UserActions extends Component {
     private renderNoActions() {
         const element = document.createElement('div');
```
lbergeron, line 318: For a moment, I was confused to see these two lines together. Didn't notice the `this.` at first.

    element.innerHTML = // the whole no actions markup...
    this.element.innerHTML = '';

I'd suggest to rename the `element` variable to `emptySpace` or something like that to avoid confusion with `this.element`. Feel free to keep the code as is if you prefer though.

src/components/UserActions/UserActions.ts
```diff
@@ -318,16 +317,33 @@ export class UserActions extends Component {
...
         element.classList.add('coveo-no-actions');
-        element.innerText = l(`${UserActions.ID}_no_actions`);
+        element.innerHTML = `
+        <div class="coveo-user-actions-title">${l(UserActions.ID)}</div>
+        <div>${l(UserActions.ID + '_no_actions_title')}.</div>
+        <br/>
```
lbergeron, line 323: Instead of adding `<br />` tags like this, you could add a class like `coveo-no-actions-section` to add the top margin you want.

tests/components/UserActions/UserActions.spec.ts
```diff
@@ -294,8 +294,10 @@ describe('UserActions', () => {
         );
         await mock.cmp.show();
 
-        expect(mock.cmp.element.querySelector<HTMLElement>('.coveo-no-actions')).not.toBeNull();
-        expect(mock.cmp.element.querySelector<HTMLElement>('.coveo-no-actions').innerText).toBe(l(`${UserActions.ID}_no_actions`));
+        const noActions = mock.cmp.element.querySelector<HTMLElement>('.coveo-no-actions');
+
+        expect(noActions).not.toBeNull();
+        expect(noActions.innerText).toContain(l(`${UserActions.ID}_no_actions_title`));
```
lbergeron, line 300: You could also validate that the possible causes are displayed in this case.

lbergeron (approved):
Only added a few minor comments. It looks good to me.

src/components/UserActions/Strings.ts
```diff
@@ -1,7 +1,13 @@
 import { Translation, Language } from '../../utils/translation';
 
 Translation.register(Language.English, {
-    UserActions_no_actions: 'No actions available for this user',
+    UserActions: 'User Actions',
+    UserActions_no_actions_title: 'No actions available for this user',
```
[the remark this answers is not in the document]
erocheleau: I should've specified :D the `.` is added but elsewhere hehe thanks for confirming I need a `.` though!

src/components/UserActions/UserActions.ts
erocheleau: Yea good idea, I was confused about it too (not my code the element part ;) )

src/components/UserActions/UserActions.ts
erocheleau: I wrote a comment but it disappeared :( 

Basically the thing we are displaying I consider more as a 1 paragraph than multiple HTML elements. What do you think? It felt more like a line break, than a margin between elements.

mikegron (approved):
tes en feu

PR #108, opened 2022-02-08
feat(sfint-4354) custom events filtering

erocheleau:
#nofilter
image

#withfilter 
image

#withallactionsfiltered
image

Don't think the third case is actually possible but I thought I still needed to include it.

### DIFF
--- a/src/components/UserActions/Strings.ts
+++ b/src/components/UserActions/Strings.ts
@@ -3,7 +3,7 @@ import { Translation, Language } from '../../utils/translation';
 Translation.register(Language.English, {
     UserActions: 'User Actions',
     UserActions_no_actions_title: 'No actions available for this user',
-    UserActions_no_actions_causes_title: 'Potential causes',
+    UserActions_no_actions_causes_title: 'Possible causes',
     UserActions_no_actions_cause_not_enabled: 'User actions are not enabled for your organization',
     UserActions_no_actions_cause_not_associated: 'There are no user actions associated with the user',
     UserActions_no_actions_cause_case_too_old: 'The case is too old to detect related actions',
--- a/src/components/UserActions/Strings.ts
+++ b/src/components/UserActions/Strings.ts
@@ -23,6 +23,8 @@ Translation.register(Language.English, {
     UserActivity_duration: 'Duration',
     UserActivity_other_event: 'Other Event',
     UserActivity_other_events: 'Other Events',
+    UserActivity_no_actions_timeline: 'No actions to display in the timeline',
+    UserActivity_no_actions_cause_filtered: 'All the actions were filtered',
 
     UserActivity_search: 'Query',
     UserActivity_query: 'User Query',
--- a/src/components/UserActions/UserActivity.ts
+++ b/src/components/UserActions/UserActivity.ts
@@ -107,12 +107,6 @@ export class UserActivity extends Component {
         this.userProfileModel = get(this.root, UserProfileModel) as UserProfileModel;
 
         this.userProfileModel.getActions(this.options.userId).then((actions) => {
-            const FAKE_EVENT_CUSTOM = new UserAction(UserActionType.Custom, new Date(), {
-                event_type: 'ticket_field_update',
-                event_value: 'caseDetach',
-                origin_level_1: 'foo',
-            });
-            actions = [FAKE_EVENT_CUSTOM];
             const sortMostRecentFirst = (a: UserAction, b: UserAction) => b.timestamp.getTime() - a.timestamp.getTime();
             const sortedActions = actions.sort(sortMostRecentFirst);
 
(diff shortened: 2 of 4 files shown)


erocheleau:
@acarpentier01 Look at the last image, this is the strings to review!

They're all displayed with a `.` 😄

src/components/UserActions/UserActivity.ts
```diff
@@ -59,6 +66,10 @@ export class UserActivity extends Component {
     static readonly ID = 'UserActivity';
     static readonly options: IUserActivityOptions = {
         userId: ComponentOptions.buildStringOption({ required: true }),
+        customActionsExclude: ComponentOptions.buildListOption({
+            defaultValue: ['ticket_create_start', 'ticket_field_update', 'ticket_next_stage', 'ticket_classification_click'],
```
tedre191, line 70: Could be extracted in an export const instead for the list of defaultValues, depends on what you prefer.

src/components/UserActions/UserActivity.ts
```diff
@@ -97,9 +108,13 @@ export class UserActivity extends Component {
 
         this.userProfileModel.getActions(this.options.userId).then((actions) => {
             const sortMostRecentFirst = (a: UserAction, b: UserAction) => b.timestamp.getTime() - a.timestamp.getTime();
-
             const sortedActions = actions.sort(sortMostRecentFirst);
-            this.sessions = this.splitActionsBySessions(sortedActions);
+
+            let filteredActions = sortedActions;
+            if (this.options.customActionsExclude && this.options.customActionsExclude.length > 0) {
+                filteredActions = actions.filter((action) => this.filterActions(action));
```
tedre191, line 115: actions.filter((action) => this.filterActions(action)); or
actions.filter(this.filterActions.bind(this)) might be clearer?

tedre191 (approved):
GG

src/components/UserActions/UserActivity.ts
lbergeron: Why not this?

```suggestion
                filteredActions = actions.filter(this.filterActions);
```

src/components/UserActions/UserActivity.ts
```diff
@@ -97,9 +108,13 @@ export class UserActivity extends Component {
 
         this.userProfileModel.getActions(this.options.userId).then((actions) => {
             const sortMostRecentFirst = (a: UserAction, b: UserAction) => b.timestamp.getTime() - a.timestamp.getTime();
-
             const sortedActions = actions.sort(sortMostRecentFirst);
```
lbergeron, line 111: Since `.sort()` sorts the elements in-place, the `actions` argument is being altered in the process. I'd suggest to spread to a new array before sorting the entries.

```suggestion
            const sortedActions = [...actions].sort(sortMostRecentFirst);
```

lbergeron (approved):

src/components/UserActions/UserActivity.ts
erocheleau: meh, they are nice and close to where they are actually used, it's fine to me I think?

src/components/UserActions/UserActivity.ts
```diff
@@ -116,6 +131,19 @@ export class UserActivity extends Component {
...
+        const eventType = action.raw.event_type || '';
+        if (action.type === UserActionType.Custom) {
+            return (
+                action.type === UserActionType.Custom &&
+                !this.options.customActionsExclude.includes(eventValue) &&
+                !this.options.customActionsExclude.includes(eventType)
+            );
+        }
+        return true;
+    }
```
louis-bompart, line 145: `action.type === UserActionType.Custom` is evaluated twice, the l139 one is redundant and will always return true.
I would split out to keep things tidy IMHO.

```suggestion
    private filterActions(action: UserAction): boolean {
        return action.type !== UserAction.Custom || this.isCustomActionExcluded(action)
    }

    private isCustomActionExcluded(action: UserAction): boolean {
        const eventValue = action.raw.event_value || '';
        const eventType = action.raw.event_type || '';
        return 
            this.options.customActionsExclude.includes(eventValue) &&
            this.options.customActionsExclude.includes(eventType);
    }
```

src/components/UserActions/Strings.ts
```diff
@@ -23,6 +23,8 @@ Translation.register(Language.English, {
     UserActivity_duration: 'Duration',
     UserActivity_other_event: 'Other Event',
     UserActivity_other_events: 'Other Events',
+    UserActivity_no_actions_timeline: 'No actions to display in the timeline',
```
acarpentier01, line 26: Potential causes-->Possible causes

acarpentier01 (approved):
Looks great!

src/components/UserActions/Strings.ts
erocheleau: Sure, that will change it for the other places where the error is displayed is that okay?

